PR #2, opened 2015-08-21
commit project settings file

kormic:


gsaslis:
Yeah, we need this as it modifies the file after running pod update (which is necessary after cloning the project).

PR #5, opened 2015-08-22
Compose a message is working

kormic:


gsaslis:
I am experiencing some issues with the keyboard not showing up when I click in the text field, and also with the scroll to bottom after a new message is written...

Could you please have a look?

Also, I don't think the keyboard should be automatically hidden after sending a message. Usually you would want to send many messages, right ?   ; )

kormic:
Keyboard not hiding and scroll to bottom might be a bug in iOS 9 simulator. We have to check on that later. 
Dismissing the keyboard after sending a single message is removed....

gsaslis:
Ok - please look into the potential simulator bug, and if you can confirm it's a bug, please resolve the merge conflicts so we can merge this in ;)

kormic:
@gsaslis Conflicts are resolved

gsaslis:
:) now that #9 was merged in, there are still conflicts...

kormic:
@gsaslis Conflicts resolved!

gsaslis:
ok - merging this in

kormic:
There is no email being saved for the user yet, that's why you can't login with your email. For the password I just added that check for demo purposes. Both will be implemented according to what check happens on the server side. About making the borders red, it's just for now. Either I will add popovers or alerts. 

The if is now return exists!!! :D

kormic:
@gsaslis Oh and another thing, on registration should the password be secure text? I guess I could add a button(?) to display the characters. I think this is better, isn't it?

gsaslis:
@kormic it should definitely have the characters hidden - not sure whether we need the ? button to start with.. perhaps we can add it as a new feature later?

kormic:
@gsaslis OK then....

PR #7, opened 2015-08-24
Move current assets to Assets.xcassets

rafaelks:


gsaslis:
hmmm - not very familiar with this, so not sure how to reply... 
if this is some sort of standard practice, I guess it should be merged in!

@rafaelks if there is any link you can share, so we can also learn about this, it'd be appreciated! ; )

rafaelks:
@gsaslis Since Apple added **xcassets** to Xcode, we should only use this, to support multiple image sizes (1x, 2x, 3x).

Documentation: https://developer.apple.com/library/ios/recipes/xcode_help-image_catalog-1.0/Recipe.html

gsaslis:
ok - got ya!! 

Looks good - merging this in! :+1:

PR #8, opened 2015-08-24
Add Podfile.lock to the project

rafaelks:


gsaslis:
yep - we definitely need this!

PR #33, opened 2015-08-30
Add Localizable.string file pattern for better improvement in future with multiple languages. Add validator for UITest that use method that is only available at iOS 9 and higer

felipeflorencio:


felipeflorencio:
I just add localizable.string for multiple languages improvement and add constant to prevent compile erro when trie compile using ios 8 as target

gsaslis:
looks good! ;) 
:+1:

Issue #26, opened 2015-08-28
Persistent Storage


gsaslis:
Add persistance storage to application. Either CoreData or SQLite

- [ ] Check CoreData vs SQLite
- [ ] Choose CoreData or SQLite
- [ ] Create prototype to store login credentials
- [ ] Expand Usage - Create appropriate data model and helpers

filipealva:
@gsaslis What do you think about [Realm](https://realm.io/)? I'm starting with Realm on a project and I'm enjoying it. Maybe would be nice to check CoreData vs SQLite vs Realm.

gsaslis:
@filipealva thanks for the suggestions, as well as the blog post! I haven't really looked into Core Data (@greenman18523 has) and I am not really sure what data we will have to persist locally (should we setup a call to discuss about that some time?) but....
I'm sold! :+1:  ; )

filipealva:
@gsaslis I feel happy that you liked it. About the call: I'm just starting to explore the project, don't know if I can contribute too much, but I think we can call the other guys to contribute with this topic (@rafaelks would you like to give your 2cents here?), and after discussing a little here, if necessary, we can schedule a call to decide some details. 

What other kind of data (besides the messages) do you think that we will need to persist?

rafaelks:
@filipealva Thanks your suggestion!

But, I think we really don't need to persist anything right now. I'm fine in just showing data downloaded from web services, just like desktop and web application does. There's no need to create complexity using some database in the app right now.

We really need to focus on getting/sending data via DDP before start thinking in persisting data. That's my opinion.

gsaslis:
FYI - i've already began working on the interconnection with the server through ddp.

filipealva:
@gsaslis @rafaelks totally agree!

When I commented here I was just browsing the issues, trying to get "synchronized" with the project.

rafaelks:
@gsaslis @filipealva For sure! In the future we certainly can persist data to offline access and a better user experience. :+1:

PR #49, opened 2015-10-03
Login & signup functionality to server

gsaslis:
login now works with the rocket.chat meteor server and so does signup & logout. 

Lots of dummy data has now been removed. 
Use of CoreData was also removed for now, since in #26 we decided we'll use sth else.. For version 1.0.0 we probably don't need to store anything anyway.


gsaslis:
closing this in favour of a new PR that will include some more features @kormic has been working on.

PR #52, opened 2015-10-12
Settings UI - Join Default Channels

kormic:
Added avatarControllers group and put in there all the view controllers that will be used for the various ways of setting the avatar. Now all settings UI are completed. 

Also added subscriptions to activeUsers and userData collections. Subscription to userData is to get the username and use it as a placeholder in profile settings. 

Lastly, now the user joins by deafult the default channels after login.

Please review and merge if ok...


gsaslis:
yep - looks good overall

PR #55, opened 2015-11-04
Getting/sending messages from/to the GENERAL room

kormic:
As I said in the previous (closed now) PR, I am not able to create a collection from the client side (using the objectiveDDP library) so I ended up subscribing to messages collection (this is a publication I found on the server code) with the room id and this gives us access to the incoming messages. After logging in the user joins the default channels and gets the 50 last messages. Then I add an observer so we can be notified every time a new message is coming in. The user can also send messages. I imported the swiftyJSON library to handle any JSON. I also added the RoomHistoryManager.swift so later we can implement the server's RoomHistoryManager although there is the possibility we won't need to.

Now we can get the GENERAL's room messages and talk to each other! :)

Please feel free to review!

A notice about the incoming messages:
When we subscribe to the messages collection, meteor puts some messages in the messages collection starting from the newest one and going on to older messages. This is not very handy when populating the tableview. So the observer is added to a later point where the messages have been already added in the messages collection and we won't get notifications for these messages. Instead I call the loadHistory in viewWillAppear to get the 50 last messages and populate the tableview and then from there incoming messages are being handled properly.

### DIFF
--- a/Rocket.Chat.iOS/ViewControllers/ViewController.swift
+++ b/Rocket.Chat.iOS/ViewControllers/ViewController.swift
@@ -30,8 +30,8 @@ class ViewController: UIViewController, UITableViewDataSource, UITableViewDelega
     //JSON to keep the response
     var chatMessages:JSON = []
     
-    //Dictionary to keep the users from the JSON response
-    var users = [Int():[String(),String(),String(),String()]]
+    //Dictionary to keep the chatMessageData from the JSON response
+    var chatMessageData = [Int():[String(),String(),String(),String()]]
     
     //Dictionary to keep the timestamps from the JSON response
     var ts = [Int():[Double()]]
@@ -86,7 +86,7 @@ class ViewController: UIViewController, UITableViewDataSource, UITableViewDelega
                 
                 
 //                                print(self.chatMessages)
-                //                print(self.users!)
+                //                print(self.chatMessageData!)
                 
                 
                 var i = self.chatMessages.count
@@ -99,7 +99,7 @@ class ViewController: UIViewController, UITableViewDataSource, UITableViewDelega
                     }
                     
                     
-                    self.users[i] = [subJson["u","_id"].string!, subJson["u","username"].string!, subJson["msg"].string!,type]
+                    self.chatMessageData[i] = [subJson["u","_id"].string!, subJson["u","username"].string!, subJson["msg"].string!,type]
                     let timestamp = [subJson["ts","$date"].number!]
                     let timestampInDouble = timestamp as! [Double]
                     self.ts[i] =  [timestampInDouble[0] / 1000]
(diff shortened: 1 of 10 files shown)


Rocket.Chat.iOS/ViewControllers/ViewController.swift
```diff
@@ -21,20 +22,30 @@ class ViewController: UIViewController, UITableViewDataSource, UITableViewDelega
...
     var dateFormatter = NSDateFormatter()
     
-    
     var meteor: MeteorClient!
     
+    //JSON to keep the response
+    var chatMessages:JSON = []
+    
+    //Dictionary to keep the users from the JSON response
+    var users = [Int():[String(),String(),String(),String()]]
```
gsaslis: Would it be possible to replace these dictionaries with an object?  It's really hard to understand what information will get stored here... ; )

Rocket.Chat.iOS/ViewControllers/ViewController.swift
```diff
@@ -42,10 +53,79 @@ class ViewController: UIViewController, UITableViewDataSource, UITableViewDelega
...
+                
+                for (_,subJson) in self.chatMessages {
+                    
+                    var type = ""
+                    if subJson["t"].string != nil {
+                        type = subJson["t"].string!
+                    }
+                    
+                    
+                    self.users[i] = [subJson["u","_id"].string!, subJson["u","username"].string!, subJson["msg"].string!,type]
```
gsaslis: I'm missing sth... Not sure why you're storing chatMessages in a structure called "users"

kormic:
@gsaslis To be honest I need to rename those variables better....Each message that is coming in has various info like user id, username, timestamp, message type etc...So probably the right name would be messageData or something like that....

Rocket.Chat.iOS/ViewControllers/ViewController.swift
kormic: I am using these structure to sort the messages in a reverse order (from oldest to newest) so I can access them easier using the indexPath when the tableview is being populated. First I need to get the response as JSON and then sort it inside this users dictionary. Maybe I could comment this better so it can be clear to others? 
Something like:
[index: [id,username,message,msgType]]

kormic:
@gsaslis I renamed the dictionary and I added a comment to explain it's structure. If it's better to change the dicts with objects then I have to think through all the logic again so please let me know....

gsaslis:
well, imho, this: 
[index: [id,username,message,msgType]]

has no reason to be a dictionary / map ... it should be an (indexed, if must be, but probably simply an ordered) collection of, let's say 'ChatMessage' objects... 
Your ChatMessage object would then have an id, username, message, msgType attributes (and any other relevant info, such as timestamps, etc. )
You can then iterate the collection and display the messages in the order you decide...  

That's the main idea, but I'm happy to provide more detailed info if u need it ; )

kormic:
@gsaslis This is EXACTLY what I am doing right now! :)
I might have one question though but I won't ask yet....

kormic:
@gsaslis Ok, please have a look and tell me if this looks better.

gsaslis:
Looks good now - merging this in, so we can continue! ;) :+1:

PR #56, opened 2015-11-15
Network Status

kormic:
This is to keep the user informed about the network status. When loosing the connection a message appears at the top with a spinning indicator. If we loose connection while we are in a chatroom then a message appears at the top, the text input changes to grey background and becomes uneditable and the send button is being disabled.

Please feel free to review...

p.s.>Should this also happen at the register and the forgot password screens?

Note: If you run this on a iphone 5 and earlier, there seems to be a bug. If I turn off my laptop's wifi the simulator doesn't seem to "see" it immediately. If I go to safari (in the simulator) and play around with my laptop's wifi then the simulator "sees" it and after this the app recognises the network connection correctly.

### DIFF
--- a/Rocket.Chat.iOS/AppDelegate.swift
+++ b/Rocket.Chat.iOS/AppDelegate.swift
@@ -55,11 +55,48 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
   
   func reportConnection() {
     print("================> connected to server!")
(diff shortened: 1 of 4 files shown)


gsaslis:
@kormic same here dude... please feel free to merge this in yourself, as I'm very pushed for time these days

PR #57, opened 2015-11-19
Setting User's Status (Working and Synced with the web client)

kormic:
This is so the user can set his status. What I 've done is this:
User can choose his status from the left menu. When the user press home or switches to another app the status remains the same. But when the user locks the device the status is set to away. When he unlocks, the status is set back to the previous status.If he puts the app to the background and then locks the device the the status stays the same. Also when the status is changed from the web client the app get's notified and the status is synced to the device. I added an observer to AccountBarViewController but I noticed that it gets plenty notifications so every time we close the left menu the observer is removed to keep network traffic low. It's being added every time we open the left menu. If it's needed later it won't be removed at all.

Please feel free to review and if anyone thinks this is not the way the status should be handled on locking the device please let me know! :) :) :)

P.s. > I would prefer this to be confirmed after my previous pr has been confirmed first. I don't think there would be any merge conflict anyway but just to be sure....


gsaslis:
@kormic haven't had the time to fully study the code here, but from a quick look this looks ok. Please feel free to merge this in, so you can proceed with the rest!

PR #58, opened 2015-12-12
Room selection in left menu and changed the subscription to stream-messages

kormic:
Now the channels in the left menu are showing. You can switch between rooms. When entering the settings now you can go back by selecting a room or by selecting a status. When you close the app completely it remembers the last room you have joined and next time loads that channel again. When you switch to a different room it scrolls down to the latest message. For now when you start the app and it loads the room give it some time to load it and then change room. If you change to another room before it loads the current room the app crashes. This will be fixed by adding a loading screen.

Also after talking with @Sing-Li I switched the subscription from messages to stream-messages in order to get the chat messages.

Please feel free to review and provide any feedback...


kormic:
Thanks for this great feedback! Makes me better!!! :smiley: Tomorrow I 'll push some commits for all these.

PR #85, opened 2016-07-17
Warn devs of a couple of gotchas in setup

Sing-Li:


README.md
```diff
@@ -15,11 +15,11 @@ Please note we are still in very early stages development, so expect lots of bug
 
 In case you're interested in playing around with the code or giving something back, here are some instructions on how to set up your project:
 
-1. Install [CocoaPods](https://cocoapods.org)
+1. Install [CocoaPods](https://cocoapods.org)  (note that you will need to install at least Ruby 2.2.3 for this to work)
```
rafaelks: Actually, you can install CocoaPods with Ruby 2.0.0+, right?

Sing-Li: For sure.  But there is a bug in 2.2.0 (default with OSX 10.11.5) that will prevent your workflow form working.  And require minimum update to 2.2.3     Got me for hours (hence the gotcha warning).

rafaelks: Sounds good, thank you @Sing-Li!

PR #171, opened 2016-11-02
update order category subscription name

CodeBanBan:
update order category subscription name to same desktop version


rafaelks:
@CodeBanBan Thank you! 👍 🚀

PR #202, opened 2016-12-11
Bump Xcode version and move build logic to a separated file

rakaramos:
@RocketChat/iOS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: objective-c
-osx_image: xcode8
+osx_image: xcode8.1
 rvm: 2.2.2
 before_install:
 - |
     gem install cocoapods -v '1.0.1' --no-rdoc --no-ri --no-document --quiet
     gem install xcpretty -N --no-ri --no-rdoc
     pod repo update 
-script: xcodebuild clean test -workspace Rocket.Chat.xcworkspace -scheme Rocket.Chat -destination "platform=iOS Simulator,name=iPhone 7" -enableCodeCoverage YES | xcpretty
+script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: objective-c
-osx_image: xcode8
+osx_image: xcode8.1
```
rafaelks: Now that we have Xcode 8.2, should we update it? :-)

Scripts/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+xcodebuild clean test -workspace Rocket.Chat.xcworkspace -scheme Rocket.Chat -destination "platform=iOS Simulator,name=iPhone 7" -enableCodeCoverage YES | xcpretty
```
rafaelks: What if we want to add separated process to test iPhone 7+ and SE? Can we just add another script call here?

rakaramos: Yes, we can have all the fun that bash scripts allows us 🎉

.travis.yml
rakaramos: If you look at [travis docs](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version), they only support Xcode 8.2 beta... I would say to wait until travis fully adopt it.

rafaelks: Docs are updated now, we can use the new version :-)

rakaramos: :tada:

rakaramos:
@rafaelks updated

rafaelks:
@rakaramos Thank you! 🚀

PR #210, opened 2016-12-28
SwiftLint

### LINKED ISSUE #205: Add SwiftLink to the project
https://github.com/realm/SwiftLint

rakaramos:
@rafaelks

@RocketChat/iOS

Closes #205


laggedHero:
As I understands WIP stands for `Work In Progress`. I've just added the right label.

rakaramos:
thanks @laggedHero, I won you a Dolly's bottle

Rocket.Chat/Controllers/Chat/ChatDataController.swift
```diff
@@ -32,107 +32,107 @@ struct ChatData {
...
+    // swiftlint:disable function_body_length cyclomatic_complexity
     func insert(_ items: [ChatData]) -> [IndexPath] {
         var indexPaths: [IndexPath] = []
         var newItems: [ChatData] = []
         var lastObj = data.last
-        var identifiers: [String] = []
-        
+        var identifiers: [String] = items.map { $0.identifier }
+
         for obj in items {
```
rafaelks: This code can be removed, right? :-)

Rocket.Chat/Controllers/Chat/ChatDataController.swift
```diff
@@ -32,107 +32,107 @@ struct ChatData {
...
                     if insert {
+                        guard let calendar = NSCalendar(calendarIdentifier: .gregorian) else { continue }
                         let date = newObj.timestamp
-                        let calendar = NSCalendar(calendarIdentifier: .gregorian)!
-                        let components = calendar.components([.day , .month, .year ], from: date)
-                        let newDate = calendar.date(from: components)
-                        let separator = ChatData(type: .daySeparator, timestamp: newDate!)!
+                        let components = calendar.components([.day, .month, .year], from: date)
+                        guard let newDate = calendar.date(from: components) else { continue }
+                        guard let separator = ChatData(type: .daySeparator, timestamp: newDate) else { continue }
```
rafaelks: If we call `continue` here, the new message will not be inserted in the list. Maybe let's insert this logic in a separated block, just to make sure that the new message get inserted?

Rocket.Chat/Managers/Socket/Response/SocketHandlers.swift
```diff
@@ -13,71 +13,70 @@ import SwiftyJSON
...
-        case .Connected: return handleConnectionMessage(result, socket: socket)
-        case .Ping: return handlePingMessage(result, socket: socket)
-        case .Changed, .Added, .Removed: return handleModelUpdates(result, socket: socket)
-        case .Error, .Updated, .Unknown: break
+
+        switch result.msg {
+            case .some(.Connected): return handleConnectionMessage(result, socket: socket)
+            case .some(.Ping): return handlePingMessage(result, socket: socket)
+            case .some(.Changed), .some(.Added), .some(.Removed): return handleModelUpdates(result, socket: socket)
+            case .some(.Error), .some(.Updated), .some(.Unknown): fallthrough
```
rafaelks: `fallthrough` will prevent us from calling the callback with the error, right? Maybe a `break` is better for now, since we only have one callback.

Rocket.Chat/Managers/Socket/Response/SocketHandlers.swift
```diff
@@ -13,71 +13,70 @@ import SwiftyJSON
...
         }
-        
-        switch result.msg! {
-        case .Connected: return handleConnectionMessage(result, socket: socket)
-        case .Ping: return handlePingMessage(result, socket: socket)
-        case .Changed, .Added, .Removed: return handleModelUpdates(result, socket: socket)
-        case .Error, .Updated, .Unknown: break
+
+        switch result.msg {
+            case .some(.Connected): return handleConnectionMessage(result, socket: socket)
```
rafaelks: I think it's better to make sure `msg` exists instead of using `.some()` to all options. What do you think?

Rocket.Chat/Managers/Socket/Response/SocketResponse.swift
```diff
@@ -10,41 +10,38 @@ import Foundation
...
     }
-    
-    
+
     // MARK: Checks
-    
+
     func isError() -> Bool {
-        if msg == .Error || result["error"] != nil {
+        if msg == .Error || result["error"] != JSON.null {
```
rafaelks: I haven't tested yet, but I believe in this case, `nil` will not be same as `JSON.null`. Have you tried it?

Rocket.Chat/Models/User.swift
```diff
@@ -11,10 +11,7 @@ import RealmSwift
 import SwiftyJSON
 
 enum UserStatus: String {
-    case offline = "offline"
-    case online = "online"
-    case busy = "busy"
-    case away = "away"
+    case offline, online, busy, away
```
rafaelks: 💯

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -13,11 +13,11 @@ protocol ChatMessageCellProtocol: ChatMessageURLViewProtocol, ChatMessageVideoVi
 }
 
 class ChatMessageCell: UICollectionViewCell {
-    
+
     static let minimumHeight = CGFloat(55)
     static let identifier = "ChatMessageCell"
 
-    var delegate: ChatMessageCellProtocol?
+    weak var delegate: ChatMessageCellProtocol?
```
rafaelks: 🤘

Rocket.Chat/Views/Chat/ChatPreviewModeView.swift
```diff
@@ -8,21 +8,21 @@
 
 import UIKit
 
-protocol ChatPreviewModeViewProtocol {
+protocol ChatPreviewModeViewProtocol: class {
```
rafaelks: 👍

Rocket.Chat/Views/Chat/ChatPreviewModeView.swift
```diff
@@ -8,21 +8,21 @@
...
     func userDidJoinedSubscription()
 }
 
 class ChatPreviewModeView: BaseView {
-    
-    var delegate: ChatPreviewModeViewProtocol?
+
+    weak var delegate: ChatPreviewModeViewProtocol?
```
rafaelks: :metal:

Rocket.Chat/Views/Sections/SubscriptionSearchMoreView.swift
```diff
@@ -8,20 +8,20 @@
...
+protocol SubscriptionSearchMoreViewDelegate: class {
     func buttonLoadMoreDidPressed()
 }
 
 class SubscriptionSearchMoreView: BaseView {
-    
-    var delegate: SubscriptionSearchMoreViewDelegate?
-    
+
+    weak var delegate: SubscriptionSearchMoreViewDelegate?
```
rafaelks: :metal:

Rocket.Chat/Controllers/Chat/ChatDataController.swift
rakaramos: 👍

Rocket.Chat/Managers/Socket/Response/SocketHandlers.swift
rakaramos: `fallthrough` will make the current clause jump to the next one, in this case:

```swift
case .some(.Error), .some(.Updated), .some(.Unknown): fallthrough
```

will resolve to:

```swift
default: break
```

But, from a maintenance perspective, if we add a new case value to the enum, the complier won't help us to keep track of the places we need to update.
So, I agree and will change from `fallthrough` to `break`

Rocket.Chat/Managers/Socket/Response/SocketResponse.swift
rakaramos: It's a warning from the library, it should not break anything 😄

Rocket.Chat/Controllers/Chat/ChatDataController.swift
rafaelks: What do you think about this code? Separating that context in another block we can use return and the message will be inserted anyway :-)

```swift
...
        func insertDaySeparator(from obj: ChatData) {
            guard let calendar = NSCalendar(calendarIdentifier: .gregorian) else { return }
            let date = obj.timestamp
            let components = calendar.components([.day, .month, .year], from: date)
            guard let newDate = calendar.date(from: components) else { return }
            guard let separator = ChatData(type: .daySeparator, timestamp: newDate) else { return }
            identifiers.append(separator.identifier)
            newItems.append(separator)
        }

        for newObj in items {
            if let lastObj = lastObj {
                if lastObj.type == .message && (
                    lastObj.timestamp.day != newObj.timestamp.day ||
                    lastObj.timestamp.month != newObj.timestamp.month ||
                    lastObj.timestamp.year != newObj.timestamp.year) {

                    // Check if already contains some separator with this data
                    var insert = true
                    for obj in data.filter({ $0.type == .daySeparator }) {
                        if lastObj.timestamp.day == obj.timestamp.day &&
                           lastObj.timestamp.month == obj.timestamp.month &&
                           lastObj.timestamp.year == obj.timestamp.year {
                            insert = false
                        }
                    }

                    if insert {
                        insertDaySeparator(from: newObj)
                    }
                }
            }

            newItems.append(newObj)
            lastObj = newObj
        }
...
```

rafaelks:
This is looking very good! Thank you @rakaramos! 👍

rakaramos:
🎉

PR #217, opened 2017-01-11
Fixed: user status offline must be updated too

### LINKED ISSUE #214: User status not updated properly
Your Rocket.Chat version: demo

Hi @rafaelks ,

the user status is not synced properly, can you recheck this?

Here are some screenshots: (made on the same time)

(User status on my iPhone, pretty much users online on the demo :-D)

When I look into my electron client.. and .. surprising :( not that much users are online:

bildschirmfoto 2017-01-08 um 22 03 28

rafaelks:
@RocketChat/iOS

Closes #214.

### DIFF
--- a/Rocket.Chat/Managers/Launcher/UserCoordinator.swift
+++ b/Rocket.Chat/Managers/Launcher/UserCoordinator.swift
@@ -1,0 +1,29 @@
+//
+//  UserCoordinator.swift
+//  Rocket.Chat
+//
+//  Created by Rafael Kellermann Streit on 10/01/17.
+//  Copyright © 2017 Rocket.Chat. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+class UserCoordinator: LauncherProtocol {
+
+    func prepareToLaunch(with options: [UIApplicationLaunchOptionsKey : Any]?) {
+        turnAllUsersOffline()
+    }
+    
+    // This method is suppose to be called when the app starts. We want all
+    // users to be offline, because when we subscribe to the API that returns
+    // the active users, it will only return users that aren't offline.
+    private func turnAllUsersOffline() {
+        // FIXME: Remove the Realm dependency.
+        Realm.execute { (realm) in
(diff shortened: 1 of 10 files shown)


rafaelks:
@laggedHero @rakaramos Can you guys review the PR? I've a few ideas to start changing the architecture to "clean", but I think we should not do it with this PR. For this one (since is a bug fix) let's focus on the bug fix only.

Rocket.Chat/Managers/Socket/Response/SocketHandlers.swift
```diff
@@ -78,8 +78,19 @@ extension SocketManager {
...
+                }
+
+                if msg == .Removed {
+                    let user = Realm.getOrCreate(User.self, primaryKey: identifier, values: fields)
+
+                    Realm.execute({ (realm) in
+                        user.status = .offline
+                        realm.add(user, update: true)
+                    })
+                }
```
rafaelks: @rakaramos @laggedHero First change I'll made is remove the Realm dependency from this file, creating the initial interactions with DB domain to update the objects.

Rocket.Chat/Managers/Launcher/UserCoordinator.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    // This method is suppose to be called when the app starts. We want all
+    // users to be offline, because when we subscribe to the API that returns
+    // the active users, it will only return users that aren't offline.
+    private func turnAllUsersOffline() {
+        // FIXME: Remove the Realm dependency.
+        Realm.execute { (realm) in
+            let users = realm.objects(User.self)
+            users.setValue("offline", forKey: "privateStatus")
+        }
```
rafaelks: Once the interactors with DB are ready, we can remove this dependency too.

laggedHero:
I think it's fine. Moving the dependency on Realm "out" will take time. Let's make sure to not slow down too much because of that.

:)

PR #225, opened 2017-01-16
update so that users can change to other servers

devinakumar:
right now, if a user logs into the mobile console, then logs out on a desktop (log out of all other devices), then there is no way for the user to make it back to the server in the mobile iOS instance.  enabling the button at all times gives people the option of being able to connect to a different server whenever they wish.

@RocketChat/iOS

Closes #ISSUE_NUMBER

### DIFF
--- a/Rocket.Chat/Controllers/MainViewController.swift
+++ b/Rocket.Chat/Controllers/MainViewController.swift
@@ -27,7 +27,7 @@ final class MainViewController: BaseViewController {
 
         if let auth = AuthManager.isAuthenticated() {
             labelAuthenticationStatus.text = "Logging in..."
-            buttonConnect.isEnabled = false
+            buttonConnect.isEnabled = true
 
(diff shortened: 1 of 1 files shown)


rafaelks:
@devinakumar Hey, thanks for the PR!

I'm sure I got your use case. You mean that, when you login into your desktop, the mobile app gets logged out? It shouldn't happen, if that's the case.

devinakumar:
I meant that when I log into the mobile app and my desktop, then go to my desktop and click "log out of all devices," the mobile app comes up with a screen that says "Logging in...connect to another server."  However, the screen is frozen here, and the "connect to another server" is grayed out.  There is no way to get back to the initial login screen or choose another server.

rafaelks:
@devinakumar Gotcha. Then I believe the error is that we're not handling the error case when trying to resume by some existing token. We should not just enable the button, because it can break normal resume flow with user interception.

PR #227, opened 2017-01-20
Fixed: handle subscription updates and save it in Realm

rafaelks:


ktt9117:
@rafaelks Thanks, mate. You saved my day!

PR #259, opened 2017-02-09
Fixed the problem that URL cannot be generated if user name is not English (e.g Korean etc).

gradler:
@RocketChat/iOS

Closes #ISSUE_NUMBER


rafaelks:
@gradler Thank you! 👍

PR #299, opened 2017-02-20
Fixed scrolling if message is bigger than container layout

### LINKED ISSUE #216: Messages is not appeared
Your Rocket.Chat version: 0.48.2

When I move the channel, the messages on the channel are not visible at first, but when the keyboard is up or the hamburger menu is pressed, the messages are seen at the moment.

(1)
image

(2)
image

For channels that does not have a lot of message history, sending a message might show only the last message you sent, hiding the keyboard and then re-appearing, or you might see a list of messages you previously sent when you clicked the hamburger menu.

(3)
image

(4)
image

gradler:
@RocketChat/iOS

Closes #216


rafaelks:
@gradler Thank you!

PR #361, opened 2017-03-21
Fixed: send of empty messages

brenoxp:
@RocketChat/iOS

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -248,6 +248,7 @@ final class ChatViewController: SLKTextViewController {
     // MARK: Message
 
     fileprivate func sendMessage() {
+        if textView.text.characters.count == 0 { return }
         guard let message = textView.text else { return }
         rightButton.isEnabled = false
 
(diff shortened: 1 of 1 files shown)


rafaelks:
@brenoxp Hey, thanks for your PR! But, how were you able to send a message with 0 chars? The "Send" button should be disabled and the keyboard "Send" button too...

brenoxp:
Yes, you're right. But I was testing the app, and I was able to send empty messages on simulator pressing enter on Mac keyboard. I think it's never too much to have twice verification.

rafaelks:
Got it, makes sense.

Can you just change it to check actually the `message` value after the `guard`? And make the `if` statement not single line.

Thank you!

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -249,6 +249,11 @@ final class ChatViewController: SLKTextViewController {
 
     fileprivate func sendMessage() {
         guard let message = textView.text else { return }
+        
+        if textView.text.characters.count == 0 {
```
rafaelks: You can use `message` instead of `textView.text`. Can you change it, please?

rafaelks:
Thank you @brenoxp!

brenoxp:
You're welcome :)

PR #395, opened 2017-04-01
Refactor chat message text view and add tests

rakaramos:
@RocketChat/iOS

### DIFF
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageTextView.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageTextView.swift
@@ -7,17 +7,14 @@
 //
 
 import UIKit
-import RealmSwift
 
 protocol ChatMessageTextViewProtocol: class {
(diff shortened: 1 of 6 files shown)


Rocket.Chat/Views/Cells/Chat/ChatMessageTextView.swift
```diff
@@ -30,65 +27,66 @@ final class ChatMessageTextView: UIView {
...
     }
 
     func viewDidTapped(_ sender: Any) {
-        guard let attachment = self.attachment else { return }
-
-        Realm.execute { (_) in
-            attachment.collapsed = !attachment.collapsed
-        }
-
+        viewModel?.toggleCollpase()
```
rafaelks: 😍

rafaelks:
Looking very good! Thank you @rakaramos!

rafaelks:
@rakaramos I had to revert your PR, because it breaks the messages that uses the Text attachment. Can you take a look and submit another one? Thank you!

PR #400, opened 2017-04-03
Fix: Long press on message cell should be responded by message cell delegate

legendecas:
@RocketChat/iOS

If long pressed on text view in message cell, the gesture event will not be sent to cell's delegate correctly as follows.

### DIFF
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
@@ -100,9 +100,9 @@ final class ChatMessageCell: UICollectionViewCell {
     func insertGesturesIfNeeded() {
         if self.longPressGesture == nil {
             let gesture = UILongPressGestureRecognizer(target: self, action: #selector(handleLongPressMessageCell(recognizer:)))
-            gesture.minimumPressDuration = 1
+            gesture.minimumPressDuration = 0.5
             gesture.delegate = self
-            contentView.addGestureRecognizer(gesture)
+            self.addGestureRecognizer(gesture)
             self.longPressGesture = gesture
         }
     }
@@ -210,7 +210,7 @@ final class ChatMessageCell: UICollectionViewCell {
 extension ChatMessageCell: UIGestureRecognizerDelegate {
 
     func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRequireFailureOf otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return false
+        return true
     }
 
 }
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
@@ -210,7 +210,7 @@ final class ChatMessageCell: UICollectionViewCell {
 extension ChatMessageCell: UIGestureRecognizerDelegate {
 
     func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRequireFailureOf otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return true
+        return false
     }
 
 }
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -100,9 +100,9 @@ final class ChatMessageCell: UICollectionViewCell {
     func insertGesturesIfNeeded() {
         if self.longPressGesture == nil {
             let gesture = UILongPressGestureRecognizer(target: self, action: #selector(handleLongPressMessageCell(recognizer:)))
-            gesture.minimumPressDuration = 1
+            gesture.minimumPressDuration = 0.5
```
rafaelks: Why did you change from 1 to 0.5?

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -100,9 +100,9 @@ final class ChatMessageCell: UICollectionViewCell {
...
             gesture.delegate = self
-            contentView.addGestureRecognizer(gesture)
+            self.addGestureRecognizer(gesture)
```
rafaelks: Why you added to the `self` instead of `contentView`?

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.xib
```diff
@@ -45,7 +45,7 @@
                             <constraint firstAttribute="width" constant="35" id="Cny-FU-tGi"/>
                         </constraints>
                     </view>
-                    <textView clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="scaleToFill" scrollEnabled="NO" editable="NO" text="Text" textAlignment="natural" translatesAutoresizingMaskIntoConstraints="NO" id="jz0-wl-Wig">
+                    <textView clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="scaleToFill" scrollEnabled="NO" editable="NO" text="Text" textAlignment="natural" selectable="NO" translatesAutoresizingMaskIntoConstraints="NO" id="jz0-wl-Wig">
```
rafaelks: 👍

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
legendecas: IMO, it would be more straightforward to add gesture to `self` since the whole view should respond to the gesture.

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
legendecas: @rafaelks I found the default duration is 0.5 second. As for 1-second duration, it might be too slow to respond to the gesture.

rafaelks: Sounds good to me! Thank you!

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -210,7 +210,7 @@ final class ChatMessageCell: UICollectionViewCell {
 extension ChatMessageCell: UIGestureRecognizerDelegate {
 
     func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRequireFailureOf otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return false
+        return true
```
rafaelks: The event is working fine, but when you return `true` here, the user will need to resign the finger from screen to open the actions. Why did you change it? I think `false` works better.

legendecas: @rafaelks I did not notice that if return `true` here the user would need to resign finger from the screen to fire event. I thought this will prevent other gesture to be fired.

I will revert this one.

rafaelks:
@legendecas Thank you, looking much better this feature now. Will wait for the tests and merge it. 👍

PR #407, opened 2017-04-06
feat: properly highlight subscription cell

legendecas:
@RocketChat/iOS

PR #416, opened 2017-04-13
Revert "feat: properly highlight subscription cell"

rafaelks:
Reverts RocketChat/Rocket.Chat.iOS#407


rafaelks:
@legendecas I'm sorry, but I'll need to revert your changes... the feature isn't working as expected when user drags the chat to the right. Can you take a look and submit another PR? Thank you!

legendecas:
@rafaelks I found the highlight will be missing while massive realm model subscription events are being fired and triggering tableView data reloading on the Subscription view. I'll take a look at it. 🤔

PR #423, opened 2017-04-15
Properly highlight subscription cell

legendecas:
@RocketChat/iOS

### DIFF
--- a/Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
+++ b/Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
@@ -345,11 +345,10 @@ extension SubscriptionsViewController: UITableViewDelegate {
     }
 
     func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
-        guard let sell = cell as? SubscriptionCell else { return }
-        if let selectedSubscription = ChatViewController.sharedInstance()?.subscription {
-            if sell.subscription == selectedSubscription {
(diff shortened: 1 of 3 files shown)


legendecas:
Ok, I'll take a look.

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -330,12 +330,27 @@ extension SubscriptionsViewController: UITableViewDelegate {
...
+
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         let subscription = self.groupSubscriptions?[indexPath.section][indexPath.row]
         let controller = ChatViewController.sharedInstance()
         controller?.closeSidebarAfterSubscriptionUpdate = true
         controller?.subscription = subscription
     }
+
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        guard let sell = cell as? SubscriptionCell else { return }
```
rafaelks: `sell`? 😂 You can use `cell` here.

legendecas: `cell` is `UITableViewCell` but not `SubscriptionCell` so it did not have property `subscription`.

legendecas: @rafaelks I don't think I can write typecasted value to immutable variable `cell`.

rafaelks: Gotcha, can you change to something like `subscriptionCell` then?

PR #549, opened 2017-07-30
fix: common crypto

legendecas:
@RocketChat/ios-sdk 

- Fixed common crypto issues

- Temporarily remove `InjectionContainer` for dependencies in built-in type extensions, we should make up with a new solution to satisfy this kind of injections.


Rocket.Chat.Shared/Managers/Launcher/PersistencyCoordinator.swift
```diff
@@ -9,7 +9,8 @@
 import Foundation
 import RealmSwift
 
-struct PersistencyCoordinator: LauncherCoordinator {
+class PersistencyCoordinator: LauncherCoordinator {
```
rakaramos: if you are changing it to a `class`, make sure to mark it as final (since there's nothing subclassing it).
That will use direct dispatch, among other compilers optimizations.
What do you think?

rakaramos:
@legendecas cool thing that you are worried about dependency inversion, nice work! Although, from what I can see, it seems to be a `Service Locator` rather than a `Dependency Injection` mechanism.

Keep up the good work buddy! 💯

Rocket.Chat.Shared/Managers/Launcher/PersistencyCoordinator.swift
rafaelks: 👍

rafaelks:
I'll merge this PR to ship the first TF and then @legendecas you can work on this final adjustment in another PR. Thanks @rakaramos

PR #553, opened 2017-08-01
[SDK] iPad adjustments and new Icon to the example app

rafaelks:


Rocket.Chat.Shared/Views/Chat/Cells/ChatMessageBubbleCell.swift
```diff
@@ -80,7 +79,7 @@ class ChatMessageBubbleCell: UICollectionViewCell, MessageTextCacheManagerInject
         }
         total += ChatMessageBubbleCell.bubblePadding + ChatMessageBubbleCell.bubbleVerticalMargin
 
-        return CGSize(width: size.width, height: total)
+        return CGSize(width: width, height: total)
```
legendecas: We shall not return the `width` of the full size width here, as the message may be too short to inflate a full sized cell.

rafaelks: Good finding!

rafaelks:
@legendecas Updated the code.

PR #560, opened 2017-08-03
fix: crashes on the second or later opening of support view controller

legendecas:
@RocketChat/ios-sdk

### DIFF
--- a/Rocket.Chat.Shared/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat.Shared/Controllers/Chat/ChatViewController.swift
@@ -304,8 +304,8 @@ public class ChatViewController: SLKTextViewController, AuthManagerInjected, Soc
         textView.resignFirstResponder()
 
         collectionView?.performBatchUpdates({
-            let indexPaths = self.dataController.clear()
-            self.collectionView?.deleteItems(at: indexPaths)
+            self.dataController.clear()
+            self.collectionView?.reloadData()
         }, completion: { _ in
             CATransaction.commit()
             self.activityIndicator.stopAnimating()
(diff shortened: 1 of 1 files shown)


Rocket.Chat.Shared/Controllers/Chat/ChatViewController.swift
```diff
@@ -452,7 +452,7 @@ public class ChatViewController: SLKTextViewController, AuthManagerInjected, Soc
                 let indexPaths = self.dataController.insert(objs)
                 collectionView.insertItems(at: indexPaths)
                 if self.messageCellStyle == .bubble {
-                    collectionView.reloadItems(at: indexPaths.filter { $0.row > 0 }.map { IndexPath(row: $0.row - 1, section: $0.section) })
+                    collectionView.reloadItems(at: indexPaths.filter { $0.row > 0 }.map { IndexPath(row: $0.row - 1, section: $0.section) }.filter { !indexPaths.contains($0) })
```
rafaelks: This line is very complex... can you simplify this code or split in more than 1 line? :-)

rafaelks: Also... we don't reload the first item?

Rocket.Chat.Shared/Controllers/Chat/ChatViewController.swift
```diff
@@ -452,7 +452,14 @@ public class ChatViewController: SLKTextViewController, AuthManagerInjected, Soc
                 let indexPaths = self.dataController.insert(objs)
                 collectionView.insertItems(at: indexPaths)
                 if self.messageCellStyle == .bubble {
-                    collectionView.reloadItems(at: indexPaths.filter { $0.row > 0 }.map { IndexPath(row: $0.row - 1, section: $0.section) }.filter { !indexPaths.contains($0) })
+                    collectionView.reloadItems(at: indexPaths
+                        // Only items not the first
+                        .filter { $0.row > 0 }
```
rafaelks: Why not the first?

legendecas: You can find that we have to minus 1 to get its previous one. the first one obviously does not have a previous one.

rafaelks: Why not simply using:

```swift
collectionView.reloadItems(at: indexPaths)
```

legendecas: since this procedure is in a batch update calls, simply using reload all the paths would accessing items that not loaded yet. so there is a crash

rafaelks: 👍

PR #565, opened 2017-08-07
[NEW] Honor UI_Use_Real_Name settings to show user full name or username all over the app

### LINKED ISSUE #543: Honor first name / last name setting
Reference to issue created on main repo: https://github.com/RocketChat/Rocket.Chat/issues/7573

There is a setting to display first name and last name instead of username.  The Rocket.Chat+ doesn't yet have that feature

rafaelks:
Closes #543

## Description

**When `UI_Use_Real_Name == true`**

- Every part of the app that uses `Subscription` model will display `fname` property;
- Every part of the app that uses `User` model will display `name` property;

**When `UI_Use_Real_Name == false`**

- Every part of the app that uses `Subscription` model will display `name` property;
- Every part of the app that uses `User` model will display `username` property;

## Progress

- [x] Honor settings
- [x] Cache settings in memory to don't need to go to DB every single time
- [x] Update other parts of the code that uses settings to use the cached one
- [x] Unit tests on new methods


chrisaudet:
Thanks!  You guys are the best.

PR #572, opened 2017-08-13
[SDK] LiveChat unit tests

legendecas:
@RocketChat/ios-sdk 

covered most methods of LivechatManager, except UI related methods.

### DIFF
--- a/Rocket.Chat.SDKTests/Managers/LivechatManagerTests.swift
+++ b/Rocket.Chat.SDKTests/Managers/LivechatManagerTests.swift
@@ -1,5 +1,5 @@
 //
-//  LiveChatManagerTests.swift
+//  LivechatManagerTests.swift
 //  Rocket.Chat
 //
 //  Created by Lucas Woo on 8/13/17.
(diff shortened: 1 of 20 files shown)


Rocket.Chat.SDKTests/Managers/LivechatManagerTests.swift
```diff
@@ -12,6 +12,7 @@ import SwiftyJSON
 
 class LivechatManagerTests: XCTestCase {
 
+    // swiftlint:disable:next force_unwrapping
     let socket = WebSocketMock(url: URL(string: "http://doesnt.matter")!)
```
rafaelks: Can't you do it in a `guard` or `if let`? Let's avoid disabling swxftlint features.

legendecas: Yes, we should do avoid disabling it in production. Since we all know it should unwrap successfully or there is something wrong with it, and this is tests anyway, I think it's okay to disable it for one exception.

rafaelks: You disabled it in many places to the same code... maybe wrap the URL into a method that will return a valid URL? I think it would be much better.

Test.Shared/WebSocketMock.swift
```diff
@@ -31,6 +31,10 @@ class WebSocketMock: WebSocket {
         return mockConnected
     }
 
+    convenience init() {
+        self.init(url: URL(string: "http://doesnt.matter")!)
```
rafaelks: Can you don't use `!`? Let's make sure the URL exists otherwise we fail the test.

PR #608, opened 2017-08-30
[NEW] Add user avatar on left-sidebar, match colors with new web default colors

micheltlutz:
@RocketChat/ios

Closes #604


micheltlutz:
img_1915
img_1914
21198444_1775010962527809_460053435_o

bugs corrected

TheReal1604:
@micheltlutz Awesome. :-)

micheltlutz:
@rafaelks You See? Sorry for the last commit that broke with a conflict, but this one is ok

rafaelks:
@micheltlutz Thank you, but looks like the font color is still not matching the web version. Can you change that too?

micheltlutz:
@rafaelks Sure, no problem

micheltlutz:
@rafaelks Can I also put the avatar in the list of direct messages in this PR?

rafaelks:
@micheltlutz Nah... let's keep it out for now, that's probably going to be a setting in the future. 👍

micheltlutz:
img_1923
img_1922

Colors Changed

rafaelks:
@micheltlutz Thank you! Can you also change the colors of the status indication? Here's a reference:

screen shot 2017-09-01 at 09 38 56

micheltlutz:
@rafaelks Yes, I already have the hex colors.

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -13,6 +13,26 @@ final class SubscriptionsViewController: BaseViewController {
...
+        }
+    }
+
+    weak var avatarView: AvatarView! {
+        didSet {
+            avatarView.layer.cornerRadius = 4
+            avatarView.layer.masksToBounds = true
+            avatarView.labelInitialsFontSize = 18
+            avatarView.frame.size.width = 40.0
+            avatarView.frame.size.height = 40.00
```
rafaelks: Let's change this to something like this?

```swift
avatarView.frame = CGRect(x: 0, y: 0, width: avatarViewContainer.width, height: avatarViewContainer.height)
```

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -13,6 +13,26 @@ final class SubscriptionsViewController: BaseViewController {
 
     @IBOutlet weak var tableView: UITableView!
     @IBOutlet weak var activityViewSearching: UIActivityIndicatorView!
+    @IBOutlet weak var imageViewIcon: UIImageView!
```
rafaelks: What's this `imageViewIcon`?

micheltlutz: I used as a reference the "AutocompleteCell" that has this code and object in the .xib. I remove?

rafaelks: @micheltlutz If it's not being used, it should be removed. Make sure it doesn't crash the app because of some XIB reference please.

micheltlutz: Removed and not crash

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
micheltlutz: Size adjusted, I changed the size definition for the updateCurrentUserInformation method so I already set the image there

micheltlutz:
Update images for preview.

img_1932
img_1931

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -13,6 +13,23 @@ final class SubscriptionsViewController: BaseViewController {
 
     @IBOutlet weak var tableView: UITableView!
     @IBOutlet weak var activityViewSearching: UIActivityIndicatorView!
+    @IBOutlet weak var avatarViewContainer: AvatarView! {
+        didSet {
+            if let avatarView = AvatarView.instantiateFromNib() {
+                self.avatarView.frame = avatarViewContainer.bounds
```
rafaelks: This isn't actually changing the `self.avatarView` right? Because it's nil at this point.

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -190,9 +207,12 @@ extension SubscriptionsViewController {
         guard let user = AuthManager.currentUser() else { return }
         guard let labelUsername = self.labelUsername else { return }
         guard let viewUserStatus = self.viewUserStatus else { return }
+        guard let avatarView = self.avatarView else { return }
 
         labelUsername.text = user.displayName()
-
+        avatarView.user = user
+        avatarView.frame = CGRect(x: 0, y: 0, width: avatarViewContainer.frame.size.width, height: avatarViewContainer.frame.size.height)
```
rafaelks: We can move this line to the `didSet {}` method and remove it from here. This method is suppose to only change data.

micheltlutz: I had tried to leave it in the didset, but at this point I get nil, I'm going to see where I can put it so I've already rendered the container on the screen.

rafaelks: I leaved a comment about why is nil... you need to user only `avatarView` instead of `self.avatarView` in that line... you don't have it in your self. reference yet.

PR #618, opened 2017-09-05
[NEW] Implement sequential (grouped) messages

### LINKED ISSUE #617: [NEW] Implement sequential (grouped) messages
Some messages should be grouped together when the same user posts in sequence.

The web app makes some exceptions like images which do not merge.
I haven't investigated which other ones don't merge yet.

cardoso:
@RocketChat/ios

Closes #617

### DIFF
--- a/Rocket.Chat/Models/Message.swift
+++ b/Rocket.Chat/Models/Message.swift
@@ -33,7 +33,7 @@ enum MessageType: String {
     case roomUnarchived = "room-unarchived"
 
     var sequential: Bool {
-        let notSequential: [MessageType] = [.image, .video]
+        let notSequential: [MessageType] = [.image, .video, .userJoined]
 
         return !notSequential.contains(self)
     }
--- a/Rocket.Chat/Models/Message.swift
+++ b/Rocket.Chat/Models/Message.swift
@@ -33,11 +33,9 @@ enum MessageType: String {
     case roomUnarchived = "room-unarchived"
 
     var sequential: Bool {
-        if self == .image {
-            return false
-        }
+        let notSequential: [MessageType] = [.image, .video]
 
-        return true
+        return !notSequential.contains(self)
     }
 }
 
--- a/Rocket.Chat/Models/Message.swift
+++ b/Rocket.Chat/Models/Message.swift
@@ -33,9 +33,9 @@ enum MessageType: String {
     case roomUnarchived = "room-unarchived"
 
     var sequential: Bool {
-        let notSequential: [MessageType] = [.image, .video, .userJoined]
+        let sequential: [MessageType] = [.text, .textAttachment]
 
-        return !notSequential.contains(self)
+        return sequential.contains(self)
     }
 }
 
--- a/Rocket.Chat/Models/Message.swift
+++ b/Rocket.Chat/Models/Message.swift
@@ -31,6 +31,14 @@ enum MessageType: String {
     case subscriptionRoleRemoved = "subscription-role-removed"
     case roomArchived = "room-archived"
     case roomUnarchived = "room-unarchived"
+
+    var sequential: Bool {
+        if self == .image {
+            return false
+        }
+
+        return true
+    }
 }
 
 class Message: BaseModel {
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
@@ -61,7 +61,7 @@ final class ChatMessageCell: UICollectionViewCell {
             MessageTextCacheManager.shared.message(for: message)?.string ?? "",
             font: UIFont.systemFont(ofSize: 15),
             width: fullWidth - 62
-            ) + (sequential ? 7 : 35)
+        ) + (sequential ? 7 : 35)
 
         for url in message.urls {
             guard url.isValid() else { continue }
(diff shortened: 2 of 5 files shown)


cardoso:
| Preview (normal) | Preview (debug) |
|-|-|
|||

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -55,13 +55,13 @@ final class ChatMessageCell: UICollectionViewCell {
...
 
-    static func cellMediaHeightFor(message: Message, grouped: Bool = true) -> CGFloat {
+    static func cellMediaHeightFor(message: Message, sequential: Bool = true) -> CGFloat {
         let fullWidth = UIScreen.main.bounds.size.width
         var total = UILabel.heightForView(
             MessageTextCacheManager.shared.message(for: message)?.string ?? "",
             font: UIFont.systemFont(ofSize: 15),
             width: fullWidth - 62
-        ) + 35
+            ) + (sequential ? 7 : 35)
```
rafaelks: Can you reduce this extra espace here?

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -87,6 +87,19 @@ final class ChatMessageCell: UICollectionViewCell {
...
+    // MARK: Sequential
+    @IBOutlet weak var labelUsernameHeightConstraint: NSLayoutConstraint!
+    @IBOutlet weak var labelDateHeightConstraint: NSLayoutConstraint!
+    @IBOutlet weak var avatarContainerHeightConstraint: NSLayoutConstraint!
+
+    var sequential: Bool = false {
+        didSet {
+            avatarContainerHeightConstraint.constant = sequential ? 0 : 35
+            labelUsernameHeightConstraint.constant = sequential ? 0 : 21
+            labelDateHeightConstraint.constant = sequential ? 0 : 21
```
rafaelks: What about even not adding values to avatar/timestamp/username labels? This way we can avoid some processing into the app.

Rocket.Chat/Models/Message.swift
```diff
@@ -31,6 +31,14 @@ enum MessageType: String {
     case subscriptionRoleRemoved = "subscription-role-removed"
     case roomArchived = "room-archived"
     case roomUnarchived = "room-unarchived"
+
+    var sequential: Bool {
+        if self == .image {
+            return false
+        }
```
rafaelks: I believe the type `video` could also be used to return `false`. What do you think?

cardoso: I can add that, but videos don't even show up here. Are they implemented?

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
cardoso: Yeah, i thought about that. I'll give it a second try, but as far as i know views with height/width == 0 are ignored by the rendering pipeline.

Rocket.Chat/Models/Message.swift
rafaelks: They show up, you need to upload a video file.

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
rafaelks: Yes, they are. But we're still constructing the avatar here: https://github.com/RocketChat/Rocket.Chat.iOS/pull/618/files#diff-535132aa8b35719fb1bcbd5602ca482eR193.

cardoso: #[Prevent unnecessary computations for sequential cells](https://github.com/RocketChat/Rocket.Chat.iOS/pull/618/commits/987713cafcea9be45ee0558466ac337cb1a696f8)

Rocket.Chat/Models/Message.swift
cardoso: #[Prevent video being sequential message](https://github.com/RocketChat/Rocket.Chat.iOS/pull/618/commits/9618ce82d685d4bfa45af35073d6d7613228654c)

rafaelks:
Thank you for the changes, but looks like the usage of this changes still has lots of bugs, look at this GIF...

cardoso:
Thanks for spotting that. I'll see if i can fix it, but i think demo.rocket.chat is down :/

rafaelks:
@cardoso Yes, it is. You can use `http://rocket.chat/deploy` to deploy a new server to test, it's free now.

cardoso:
Rookie mistake, forgot to reset the property in prepareForReuse

#[Reset sequential cell property in prepareForReuse (bugfix)](https://github.com/RocketChat/Rocket.Chat.iOS/pull/618/commits/b339f846c361c131eb20c21d2be529627c8852d7)

rafaelks:
@cardoso Please, let me know when this PR is ready to review again, OK? Thanks!

cardoso:
I think it's ready. I couldn't be 100% sure which types of messages are sequential or not, but i t can be changed easily later.

rafaelks:
@cardoso Looking very good from my tests. Tested lots of different situations, with different kind of messages and different devices.

I've one suggestion, see what you think: let's also include the user information if the difference of time between one message and another is more than 30 minutes? Let me explain you why I think that... if I'm reading a channel and one user sent 5 messages, I read everything and I'm done. I come back and there's one new message from the same user, 1 hour after the others. I think adding user avatar/username in this situation will be more effective for me to read the new message. What do you think?

cardoso:
Definitely a fast tweak, I'll look into it

PR #623, opened 2017-09-07
[REFACTOR] Remove optional from ChatData initializer

cardoso:
@RocketChat/ios


cardoso:
I think this optional isn't really necessary

rafaelks:
Looking good! 👍

PR #637, opened 2017-09-11
[FIX] Reworked Markdown parsing (emulate web version's)

### LINKED ISSUE #458: Markdown on URLs
Current behaviour is to format URLs like .../download_this_fine_file.pdf like .../download _this_ fine_file.pdf (without the whitespaces). So the parser seems to be a bit too aggressive. Happening in 1.1.0 as far as I can test.

### LINKED ISSUE #634: [BUG] Markdown parsing different from Web version
On regular web frontend asterisk and underscore differentiate bold and italic text. In this app the _regular_ parsing like here on GitHub is used with one or two asterisks/underscores setting the style to italic or bold.

cardoso:
@RocketChat/ios

Closes #634
Closes #458

This PR reworks the markdown parsing to use the same logic and support all the tags the Web version does. It replaces the old parser with [RocketChat/RCMarkdownParser](http://RocketChat/RCMarkdownParser.git) where most of the work was done.

 More details [here](https://github.com/RocketChat/RCMarkdownParser/pull/2)

# Preview

img

Some markdown samples for testing can be found in this [gist](https://gist.github.com/cardoso/e0ed330dd49f5f865012224b0d65fb3f#file-rcmarkdowntest-txt)

Note: this makes the parsing accurate, rendering still has ways to go.


rafaelks:
Looking great @cardoso!

soundstorm:
Hope that this will fix emoji parsing too 👍
 e.g. the emoji  🛠 (hammer_and_wrench) gets parsed as hammer _and_ wrench (no spaces) in 1.3.0.

cardoso:
@soundstorm thanks for pointing that out. I wasn't aware of this problem and i'm not sure if it fixes that. Since it has been merged, can you test it?

I tried copying and pasting the emoji in both web and the iOS app and the iOS app receives the emoji correctly.

PR #651, opened 2017-09-13
[FIX] Update subscriptions while searching

### LINKED ISSUE #642: [BUG] Search results in the list of Subscription doesn't update when chat room alerts is cleared or updated
# Description

If user is searching by some room, group or user, the updates into the Subscription doesn't make any effect. The example below shows a DM with my test user `rafaelks.test.2`, the correct number of unread messages is **2** instead of **1.** The number didn't update because the search is enabled.

simulator screen shot 12 sep 2017 09 53 51

cardoso:
@RocketChat/ios

Closes #642 

Updates weren't happening because the controller would check if the user was searching and would pause the updates.

I removed the checks and started filtering the incoming updates.

### DIFF
--- a/Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
+++ b/Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
@@ -1,0 +1,30 @@
+//
+//  SubscriptionExtensions.swift
+//  Rocket.Chat
+//
+//  Created by Matheus Cardoso on 9/13/17.
+//  Copyright © 2017 Rocket.Chat. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
```diff
@@ -0,0 +1,30 @@
...
+
+extension Results where Element == Subscription {
+    func sortedByLastSeen() -> Results<Subscription> {
+        return self.sorted(byKeyPath: "lastSeen", ascending: false)
+    }
+
+    func filterBy(name: String) -> Results<Subscription> {
+        return self.filter("name CONTAINS %@", name)
+    }
+}
```
rafaelks: 👍

PR #653, opened 2017-09-14
[FIX] Show cancel button if search not empty + reset text when pressed

### LINKED ISSUE #645: [IMPROVEMENT] Let "Cancel" button visible after making a search
Today, if user searches by something and select a channel, the keyboard is dismissed (and that's 100%) but the "Cancel" button isn't visible too. If the text in the search field isn't blank, we need to keep the "Cancel" button visible.

cardoso:
@RocketChat/ios

Closes #645 

Also noticed the cancel button doesn't remove the text and reset the search after pressing like the web version, so i fixed that too.

### DIFF
--- a/Rocket.Chat/Controllers/Chat/MainChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/MainChatViewController.swift
@@ -54,25 +54,23 @@ class MainChatViewController: SideMenuController, SideMenuControllerDelegate {
 
     func sideMenuControllerWillHide(_ sideMenuController: SideMenuController) {
         ChatViewController.shared?.textView.resignFirstResponder()
-        SubscriptionsViewController.shared?.textFieldSearch.resignFirstResponder()
+        SubscriptionsViewController.shared?.willHide()
     }
 
     func sideMenuControllerDidHide(_ sideMenuController: SideMenuController) {
         ChatViewController.shared?.textView.resignFirstResponder()
-        SubscriptionsViewController.shared?.textFieldSearch.resignFirstResponder()
+        SubscriptionsViewController.shared?.didHide()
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Controllers/Chat/MainChatViewController.swift
```diff
@@ -54,25 +54,23 @@ class MainChatViewController: SideMenuController, SideMenuControllerDelegate {
 
     func sideMenuControllerWillHide(_ sideMenuController: SideMenuController) {
         ChatViewController.shared?.textView.resignFirstResponder()
-        SubscriptionsViewController.shared?.textFieldSearch.resignFirstResponder()
+        SubscriptionsViewController.shared?.willHide()
```
rafaelks: 👌

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -125,15 +121,42 @@ final class SubscriptionsViewController: BaseViewController {
...
 
     @IBAction func buttonCancelSearchDidPressed(_ sender: Any) {
         textFieldSearch.resignFirstResponder()
+        textFieldSearch.text = ""
     }
 
     func searchBy(_ text: String = "") {
         guard let auth = AuthManager.isAuthenticated() else { return }
-        subscriptions = auth.subscriptions.filter("name CONTAINS %@", text)
+        subscriptions = auth.subscriptions.filterBy(name: searchText)
```
rafaelks: `searchText` should be `text` here, right?

cardoso: Copy and paste is my nemesis!!

PR #663, opened 2017-09-14
[FIX] Calculate message height with AttributedString instead of raw text

### LINKED ISSUE #647: [BUG] Message's view height can be too small for some message contents
ChatMessageCell.cellMediaHeightFor needs to be reworked to take into account NSAttributedString's attributes and not just the raw message (or even completely avoided!)

This can be verified by writing any markdown that increases content size (eg. # title, bottom part will be left out.)

example

On the other hand, height can be bigger than necessary if there's too much markdown for little output: example2

cardoso:
@RocketChat/ios

Closes #647

# Preview
preview

### DIFF
--- a/Rocket.Chat/Extensions/NSAttributedStringExtensions.swift
+++ b/Rocket.Chat/Extensions/NSAttributedStringExtensions.swift
@@ -8,6 +8,17 @@
 
 import UIKit
 
+extension NSAttributedString {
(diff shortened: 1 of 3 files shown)


rafaelks:
@cardoso I guess we can also change the calculation in `ChatMessageTextView.swift`? Looking good!

PR #667, opened 2017-09-15
[NEW] Ask user for review using iOS 10 API

### LINKED ISSUE #620: [NEW] Ask user for review using iOS 10 API
New on iOS 10, the app can ask for review in the store by using a native prompt. We should implement that to increase the number of reviews in the App Store

falcaoaugustos:
@RocketChat/ios

Closes #620 

I'd the feature for ask user for review using iOS 10.3 API (SKStoreReviewController).

This method is called when user closes settings screen for test, but the right moment to do that deserves more discussion. There are some references about it below:

https://dzone.com/articles/how-and-when-to-ask-for-reviews-including-ios-103
https://medium.com/circa/the-right-way-to-ask-users-to-review-your-app-9a32fd604fca

TODO: 
- [x] Tests.


rafaelks:
@falcaoaugustos Thanks a lot for your PR! 👍

Rocket.Chat/Managers/UserReviewManager.swift
```diff
@@ -0,0 +1,56 @@
...
+            defaults.set(newDate, forKey: nextDateForReviewKey)
+        }
+    }
+
+    var availableForReview: Bool {
+        return nextDateForReview < Date()
+    }
+
+    func calculateNextDateForReview() -> Date {
+        let week: TimeInterval = 30 // 604800
```
rafaelks: You meant to be `30` or you want `604800`?

rafaelks: Also, let's add this as a constant? This way we can use the same value in our tests :-)

falcaoaugustos: The correct value is 604800, this is the time of a week in seconds. I used 30 seconds to test with simulator.

It is ok the week value be a constant.

rafaelks:
Thank you!

PR #677, opened 2017-09-18
[FIX] Temp message and text field spacing

### LINKED ISSUE #673: [BUG] The space between message & text field is different when message is temporary
ss

falcaoaugustos:
@RocketChat/ios

Closes #673

Every time a new message is appended in the collection view, a method is called. The method completion was nil and now it set a new offset value to the collection view.


rafaelks:
Looking good @falcaoaugustos, thank you!

PR #678, opened 2017-09-19
[NEW] Start using REST API

cardoso:
@RocketChat/ios

Implementing wrappers over Rocket.Chat's REST API

The current idea is we specify the kind of resource we want to receive and let generics + type inference take care of the job:

```swift
API.shared.fetch(InfoRequest()) { result in
    guard let version = result.version else {
        print("InfoRequest failed to retrieve Rocket.Chat version")
    }
    print("Rocket.Chat version is \(version)")
}
```

```swift
API.shared.fetch(LoginRequest("username", "password")) { result in
    guard let token = result.authToken else {
        print("Invalid username or password")
    }
    print("Congratulations! Your auth token is: \(token)")
}
```

All we need to do when we want to add a new API is to subclass APIRequest and extend APIResult:

```swift
typealias InfoResult = APIResult<InfoRequest>

class InfoRequest: APIRequest {
    static let path: String = "/api/v1/info"
}

extension APIResult where T == InfoRequest {
    var version: String? {
        return self.raw?["info"]["version"].string
    }
}
```

```swift
typealias LoginResult = APIResult<LoginRequest>

class LoginRequest: APIRequest {
    static let method: String = "POST"
    static let path = "/api/v1/login"

    let username: String
    let password: String

    init(_ username: String, _ password: String) {
        self.username = username
        self.password = password
    }

    func body() -> Data? {
        let string = """
        { "username": "\(username)", "password": "\(password)" }
        """

        return string.data(using: .utf8)
    }
}

extension APIResult where T == LoginRequest {
    var data: JSON? {
        return raw?["data"]
    }

    var authToken: String? {
        return data?["authToken"].string
    }

    var userId: String? {
        return data?["userId"].string
    }
}
```

### DIFF
--- a/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
@@ -121,6 +121,8 @@ final class ConnectServerViewController: BaseViewController {
             text = defaultURL
         }
 
+        API.shared.host = text
+
         guard let url = URL(string: text) else { return alertInvalidURL() }
         guard let socketURL = url.socketURL() else { return alertInvalidURL() }
         guard let validateURL = url.validateURL() else { return alertInvalidURL() }
@@ -132,7 +134,7 @@ final class ConnectServerViewController: BaseViewController {
 
         serverURL = socketURL
 
-        validate(url: validateURL) { [weak self] (_, error) in
+        validate { [weak self] (_, error) in
             guard !error else {
                 DispatchQueue.main.async {
                     self?.connecting = false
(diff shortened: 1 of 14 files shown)


Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -121,6 +121,8 @@ final class ConnectServerViewController: BaseViewController {
             text = defaultURL
         }
 
+        API.shared.host = text
```
rafaelks: This can't be used this way, `text` may not have the protocol.

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -163,39 +165,27 @@ final class ConnectServerViewController: BaseViewController {
...
-        let task = session.dataTask(with: request, completionHandler: { (data, _, _) in
-            if let data = data {
-                guard let json = try? JSON(data: data) else { return completion(nil, true) }
-                Log.debug(json.rawString())
-
-                guard let version = json["version"].string else {
-                    return completion(nil, true)
-                }
+    func validate(completion: @escaping RequestCompletion) {
+        API.shared.fetch { (result: InfoResult?) -> Void in
```
rafaelks: I'm not sure I understand how this works... `API.shared.fetch` will hit `/api/info` because the `T` is `InfoResult`? I don't feel like this is flexible and easy to understand.

Rocket.ChatTests/API/Requests/InfoRequestSpec.swift
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                }
+            }
+        """
+
+        let json = JSON(parseJSON: jsonString)
+
+        let result = InfoResult(raw: json)
+
+        XCTAssertEqual(result.version, "0.47.0-develop")
```
rafaelks: 👌

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
cardoso: You're right! I'll find a good way to handle this soon.

PR #679, opened 2017-09-19
[FIX] Add multiple instances of the same server

### LINKED ISSUE #669: [BUG] You can add multiple instances of the same server
We shouldn't be able to add multiple instances of the same server.
The expected behavior is to just redirect the user to the currently existing server.

|Image|
|-|
||

macabeus:
@RocketChat/ios

Closes #669

### DIFF
--- a/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
@@ -125,6 +125,23 @@ final class ConnectServerViewController: BaseViewController {
         guard let socketURL = url.socketURL() else { return alertInvalidURL() }
         guard let validateURL = url.validateURL() else { return alertInvalidURL() }
 
+        if let servers = DatabaseManager.servers {
+            let sameServerIndex = servers.index(where: {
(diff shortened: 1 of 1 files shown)


rafaelks:
@macabeus Thank you for your PR! You need to sign the CLA so we can merge the PR.

macabeus:
@rafaelks Ok, I made it now =]

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -125,6 +125,23 @@ final class ConnectServerViewController: BaseViewController {
...
+                    let serverUrl = URL(string: stringServerUrl) {
+
+                    return serverUrl == socketURL
+                } else {
+                    return false
+                }
+            })
+
+            if let sameServerIndex = sameServerIndex {
+                MainChatViewController.shared?.changeSelectedServer(index: sameServerIndex)
```
rafaelks: What if `MainChatViewController.shared` doesn't exists here?

macabeus: Sorry, but I don't know... What do you prefer that I do?

PR #680, opened 2017-09-20
[NEW] User list in channel

### LINKED ISSUE #285: [NEW] User list in channel
image

There is a need for a list of online / offline users who joined the channel

cardoso:
@RocketChat/ios

Closes #285

This is in early stages and is being build on top of PR #678 .

TODO:
- [x] New ViewController
- [x] Fetch channel info using REST and populate TableView
- [x] MVVM
- [x] User status and avatar
- [x] "Show more" button for pagination
- [x] Infinite Scrolling
- [x] Support Groups and DMs
- [x] Localization
- [x] Tests


Rocket.Chat/Controllers/Chat/ChatControllerAutocomplete.swift
```diff
@@ -77,7 +77,7 @@ extension ChatViewController {
         if let user = searchResult[key] as? User {
             cell.avatarView.isHidden = false
             cell.imageViewIcon.isHidden = true
-            cell.avatarView.user = user
+            cell.avatarView.username = user.username
```
rafaelks: Don't need that anymore, right? Cause we're getting the User object.

Rocket.Chat/API/Requests/Subscription/SubscriptionInfoRequest.swift
```diff
@@ -0,0 +1,61 @@
+//
```
rafaelks: I don't think this request is being used anymore.

rafaelks:
Few UI suggestions:
1. Change member cell selection style to **none**;
2. The offline color of the status is `black`, should use the default colors

https://github.com/RocketChat/Rocket.Chat.iOS/blob/79509479a4c03d77ac70bf49387b8134bb00cb07/Rocket.Chat/Extensions/UIColorExtension.swift#L24-L38

3. Remove the `SHOW MORE` button from the loading cell and add an activity indicator;
4.  Load new page way before getting to the bottom of the list;
5. Align the user avatar (left-side) with the cell separator;

screen shot 2017-09-25 at 11 07 10 am

Rocket.Chat/Models/User.swift
```diff
@@ -28,6 +28,8 @@ class User: BaseModel {
         get { return UserStatus(rawValue: privateStatus) ?? UserStatus.offline }
         set { privateStatus = newValue.rawValue }
     }
+
+    var utcOffset: Double?
```
rafaelks: Needs to add `@objc dynamic` here also, this way we can save this information in database.

cardoso: Swift doesn't let me @objc a Double 🤷‍♂️ Any ideas?

PR #686, opened 2017-09-21
[NEW] Show server options on long-press into the servers list

### LINKED ISSUE #670: [NEW] Allow user to delete servers without being logged in on it
Today you can only remove a server if you're logged in on it. We need to let the user remove the server without being logged in. A suggestion is a long press into the server cell.

rafaelks:
@RocketChat/ios

Closes #670


cardoso:
Nice! This does not work for the server I'm currently in though, only the others. Is this intended?

rafaelks:
@cardoso Yes, it is (https://github.com/RocketChat/Rocket.Chat.iOS/pull/686/files#diff-1f15a4189e8a873c6f616c42fdfb057fR55). Let's keep it simple for now.

PR #712, opened 2017-09-24
[FIX] Google Sign In race condition

### LINKED ISSUE #705: [BUG] Google "sign In" must be pressed again after choosing account (sometimes)
This bug is most likely due to a race condition. (Sometimes it happens, sometimes it doesn't)

cardoso:
@RocketChat/ios

We had a race condition between Realm and the authentication process that could cause some annoying bugs, now I'm using a REST call to retrieve the user data initially in the authentication process.

Closes #705


Rocket.Chat/API/Requests/Authentication/MeRequest.swift
```diff
@@ -0,0 +1,25 @@
...
+
+extension APIResult where T == MeRequest {
+    var user: User? {
+        guard let raw = raw else { return nil }
+
+        let user = User()
+        user.map(raw, realm: nil)
+        return user
+    }
+}
```
rafaelks: Let's write some tests for this class?

PR #721, opened 2017-09-25
[FIX] Sometimes, the buttonScrollToBottom doesn't hide

macabeus:
@RocketChat/ios


rafaelks:
@cardoso @macabeus Still happens to me...

PR #727, opened 2017-09-25
[NEW] Support quoting

### LINKED ISSUE #706: [NEW] Support "Quote" in a message
At the moment it is not possible to quote a message from another user inside of the app.

cardoso:
@RocketChat/ios

Closes #706


rafaelks:
@cardoso Working just fine, but when I send a quoted message, the quote part is hidden, I need to scroll to view the quoted one. Example:



Also, one idea to improve this feature (and the reply feature that works the same way) would be: add a view like the example below saying that user is quoting (or replying) a message instead of adding the link of the message with markdown format. It's odd if you're not in tech world.

img_1820
img_1819

PR #743, opened 2017-09-29
[NEW] Allow the creation of channels and groups

### LINKED ISSUE #707: [NEW] Allow the creation of channels and groups
private and public channels - and invite other users to.

macabeus:
@RocketChat/ios

Closes #707

- [x] Add + button to move the screen to create a new channel 
- [x] Screen to configure your new channel
- [x] Better UI, equal to mockup
- [x] Call API to create a new channel
- [x] Create private channel
- [x] Create read-only channel
- [x] Show errors messages for user
- [x] When a new channel is created, move the user for the new channel

**To fix (rafaelks) - 1**

- [x] Use native empty button of UITextField
- [x] Move files wrongly added in root
- [x] Fix tap gesture at SubscriptionsViewController
- [x] Localize strings
- [x] Change names to others more flexible/expressive

**To fix (cardoso)**

- [x] Disable the selection of cells
- [x] Prevent the user to tap twice times the Create button
- [x] Change name "room" to "subscription" (or not?)

**To fix (rafaelks) - 2**

- [x] Labels font at new channel cell
- [x] Size of cells at new channel
- [x] Open new channel view as modal on iPad

Preview


macabeus:
@rafaelks apart from the invite people to channel, I finish this feature.



I made the UI thinking to reuse the code in others parts of project that need a TableView for configuration.

Rocket.Chat/Views/ConfigTable/ConfigTableCellTextFieldCell.swift
```diff
@@ -0,0 +1,43 @@
...
+class ConfigTableCellTextFieldCell: UITableViewCell, ConfigTableCellProtocol {
+    static let identifier = "kConfigTableCellTextField"
+    static let xibFileName = "ConfigTableCellTextFieldCell"
+    static let defaultHeight: Float = 40
+    weak var delegate: ConfigTableCellDelegate?
+    var key: String?
+
+    @IBOutlet weak var imgRoomIcon: UIImageView!
+    @IBOutlet weak var textFieldInput: UITextField!
+    @IBOutlet weak var imgClearText: UIImageView!
```
rafaelks: I honestly think that this cell `ConfigTableCellTextFieldCell` having properties like `imgRoomIcon` and `imgClearText` will cause us many problems in the future. I prefer to go the simple way and just create a cell class to each specifically cell we need right now, except for the ones you see cases that you can **really** reuse. This cell is being used only once and is looks very over engineered.

Rocket.Chat/Views/ConfigTable/ConfigTableCellProtocol.swift
```diff
@@ -0,0 +1,89 @@
...
+        cell.delegate = delegate
+        cell.key = key
+
+        if let previousValue = cell.delegate?.getPreviousValue(key: key) {
+            cell.setPreviousValue(previous: previousValue)
+        }
+
+        return cell
+    }
+}
```
rafaelks: I also feel that this protocol is over engineered and not very flexible in the way you use it. Can we try to just create the cells into the Storyboard file and point to a unique class to each cell? Looks more reasonable to me for this situation.

If we really want to reuse components in the future this way, I would say to first look how [IGListKit](https://instagram.github.io/IGListKit/index.html) is doing and try to extract the same logic.

macabeus:
@rafaelks Done! I fixed the points that you said.
Now it's good, or needs more fix/improvements?

cardoso:
Hi @macabeus, thanks for your contributions!

This is a very big feature, so we need to be careful here.

I agree with @rafaelks that there's some over-engineering and we need to discuss if it's worth it and do it right, because we could reuse this for other things such as the registration form with custom fields.

I will point out a couple things I found.

Rocket.Chat/API/Requests/Room/RoomCreateRequest.swift
```diff
@@ -0,0 +1,59 @@
...
+        switch self {
+        case .channel:
+            return "/api/v1/channels.create"
+        case .group:
+            return "/api/v1/groups.create"
+        }
+    }
+}
+
+class RoomCreateRequest: APIRequest {
```
cardoso: On the API, can you use SubscriptionCreate... instead of RoomCreate... and put the files under Subscription? That's the naming we're using for the API and we don't want to introduce conflicting nomenclature.

Rocket.Chat/NewRoomViewController.swift
```diff
@@ -0,0 +1,201 @@
...
+//  Rocket.Chat
+//
+//  Created by Bruno Macabeus Aquino on 27/09/17.
+//  Copyright © 2017 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+import RealmSwift
+
+class NewRoomViewController: BaseViewController {
```
cardoso: Again, since we don't want to introduce conflicting nomenclature, can you use CreateChannel... instead of NewRoom...?

Rocket.Chat/en.lproj/Localizable.strings
```diff
@@ -56,6 +56,15 @@
 "user_menu.settings" = "My Account";
 "user_menu.logout" = "Logout";
 
+// New Room View
```
cardoso: Rename these to Create Channel

Rocket.Chat/pt-BR.lproj/Localizable.strings
```diff
@@ -56,6 +56,15 @@
 "user_menu.settings" = "Minha conta";
 "user_menu.logout" = "Sair";
 
+// New Room View
```
cardoso: Rename these to Create Channel

macabeus:
@cardoso Thank you for revision. But, I have one question about rename "room" to "subscription", because this feature is about create specifically a new room. I checked it in nodejs code:

```javascript
RocketChat.API.v1.addRoute('channels.create', { authRequired: true }, {
	post() {

		...

		return RocketChat.API.v1.success({
			channel: RocketChat.models.Rooms.findOneById(id.rid, { fields: RocketChat.API.v1.defaultFieldsToExclude })
		});
	}
});
```

Subscription, as far as I understand, is about stream a channel in order to receive updates, and, a channel could be a room. This feature is about create a new room, and not to create a new stream.

cardoso:
I agree that we have a naming problem, but let's try to keep it consistent so if we decide to change it we only have to look for a single word.

Rocket.ChatTests/API/Requests/Room/RoomCreateRequestSpec.swift
```diff
@@ -0,0 +1,41 @@
...
+        guard let bodyJson = try? JSON(data: httpBody) else {
+            return XCTFail("body is invalid json")
+        }
+
+        XCTAssertEqual(request.url?.path, "/api/v1/channels.create", "url is correct")
+        XCTAssertEqual(request.httpMethod, "POST", "http method is correct")
+        XCTAssertEqual(request.value(forHTTPHeaderField: "Content-Type"), "application/json", "content type is correct")
+        XCTAssertEqual(bodyJson["name"].string, paramRoomName, "parameter read only is correct")
+        XCTAssertEqual(bodyJson["readOnly"].bool, paramReadOnly, "read only was set as false")
+    }
```
rafaelks: 👏

rafaelks:
@macabeus Thank you! Some UI adjustments that I think we should do:

- On both iPhone/iPad, there's some fonts & spacing adjustments to be made:
img_0001

- On iPad, let's open the modal the same way we open the "My Account" modal:
img_0007
img_0006

rafaelks:
@macabeus Merged with develop and made a few changes. Please @macabeus and @cardoso help me reviewing this code, the feature seems to be working pretty fine! 👏

rafaelks:
@macabeus Thank you very much, looking good! 💯

PR #749, opened 2017-10-02
[CHORE] Use RCMarkdownParser branch with crash fixes

cardoso:
@RocketChat/ios 

We'll be using this branch until we can confirm the crashes are fixed, since they couldn't be reproduced by the dev team yet.


rafaelks:
Do we still need this PR? I merged your PR into the lib already to master.

cardoso:
@rafaelks the idea was to use the branch until the crash was fixed then I would merge into master, but yeah this PR is not needed anymore

PR #752, opened 2017-10-03
[CHORE] Use TableView instead of CollectionView for Chat

cardoso:
@RocketChat/ios

Started using TableView in ChatViewController and changed all CollectionViewCells to TableViewCells

This will make it easier to develop #345 and #346


rafaelks:
Looking great! Just a few things:

1. Left & Right spacing of date separator cell is different;
2. Left & Right spacing of header cell for DM is different;
3. Left & Right spacing of header cell for Changel/Group is different;
4. When I open a chat, I always stays in the top; Needs to go to the bottom of the screen;
5. Toolbar that contains the textView is translucent, should not be;
6. Sending a message makes some really odd animations & stop in the wrong scroll position;

rafaelks:
I'll close this one since we're not going to use anymore, but I'll let the branch opened. Thanks @cardoso!

PR #756, opened 2017-10-05
[FIX] Tolerate wrong or lack of scheme in connect server view

### LINKED ISSUE #755: [BUG] Not adding https:// by default when adding new server
This is causing the app to not be able connect if you just type an url without scheme "open.rocket.chat"

cardoso:
@RocketChat/ios

Closes #755

### DIFF
--- a/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
@@ -123,18 +123,19 @@ final class ConnectServerViewController: BaseViewController {
     }
 
     func normalizeInputURL(_ inputURL: String) -> String? {
-        if inputURL.isEmpty {
+        guard let url = URL(string: inputURL) else {
             return nil
         }
 
-        if let index = inputURL.range(of: "://")?.upperBound {
-            let url = inputURL[index..<inputURL.endIndex]
-            if url.isEmpty { return nil }
-
-            return "https://\(url)"
+        if let host = url.host, !host.isEmpty {
+            return "https://\(host)"
         }
 
-        return "https://\(inputURL)"
+        if !url.path.isEmpty {
+            return "https://\(url.path)"
+        }
+
+        return nil
     }
 
     func connect() {
--- a/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
@@ -122,13 +122,31 @@ final class ConnectServerViewController: BaseViewController {
         present(alert, animated: true, completion: nil)
     }
 
-    func connect() {
-        var text = textFieldServerURL.text ?? ""
-        if text.characters.count == 0 {
-            text = defaultURL
+    func normalizeInputURL(_ inputURL: String) -> String? {
+        if inputURL.isEmpty {
+            return nil
         }
 
-        guard let url = URL(string: text) else { return alertInvalidURL() }
+        if let index = inputURL.range(of: "://")?.upperBound {
+            let url = inputURL[index..<inputURL.endIndex]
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -122,13 +122,31 @@ final class ConnectServerViewController: BaseViewController {
...
+        }
+
+        if let index = inputURL.range(of: "://")?.upperBound {
+            let url = inputURL[index..<inputURL.endIndex]
+            if url.isEmpty { return nil }
+
+            return "https://\(url)"
+        }
+
+        return "https://\(inputURL)"
```
rafaelks: I think will be safer here if you transform the URL into URL object and check for the protocol.

Rocket.ChatTests/Controllers/ConnectServerViewControllerSpec.swift
```diff
@@ -0,0 +1,23 @@
...
+class ConnectServerViewControllerSpec: XCTestCase {
+    func testNormalizeInputURL() {
+        let vc = ConnectServerViewController()
+
+        XCTAssertEqual(vc.normalizeInputURL("open.rocket.chat"), "https://open.rocket.chat", "will add https")
+        XCTAssertEqual(vc.normalizeInputURL("https://open.rocket.chat"), "https://open.rocket.chat", "will return correct url")
+        XCTAssertEqual(vc.normalizeInputURL("http://open.rocket.chat"), "https://open.rocket.chat", "will force https scheme")
+        XCTAssertNil(vc.normalizeInputURL("https://"), "will return nil when hostless")
+        XCTAssertNil(vc.normalizeInputURL(""), "will return nil when empty")
+    }
```
rafaelks: 👏

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
cardoso: Will do that, but it's kinda awkward because of this behavior:
https://stackoverflow.com/questions/35960185/nsurlcomponents-builds-scheme-without

URL("open.rocket.chat") means host = nil && path = "open.rocket.chat" 🤣

rafaelks: 😱

Rocket.ChatTests/Controllers/ConnectServerViewControllerSpec.swift
rafaelks: I would also include a test here that verifies if the path keeps on the URL and some query parameters:

- `https://open.rocket.chat/chat`
- `open.rocket.chat/chat`
- `open.rocket.chat/chat/foo`
- `open.rocket.chat/chat/foo?test=1`
- `open.rocket.chat/chat/foo?test=1&test=2`

What do you think?

cardoso: They actually don't. I didn't realize that was the intended behavior.

rafaelks: @cardoso Yes, there's some cases when you can use path and maybe some query URL? :-)

rafaelks: @cardoso You can use this method that I created long time ago to generate the URL in the API abstraction...

https://github.com/RocketChat/Rocket.Chat.iOS/blob/2d1eea6f62310d274713b9568dd7a97181f5f23c/Rocket.Chat/Extensions/URLExtension.swift#L26-L34

rafaelks: You can see that I even consider the port here ^^

PR #758, opened 2017-10-06
[FIX] Scroll button not hiding

### LINKED ISSUE #757: [BUG] "Scroll to bottom" button not hiding when in bottom

cardoso:
@RocketChat/ios

Closes #757

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -118,6 +118,11 @@ final class ChatViewController: SLKTextViewController {
         view.bringSubview(toFront: activityIndicatorContainer)
         view.bringSubview(toFront: buttonScrollToBottom)
         view.bringSubview(toFront: textInputbar)
+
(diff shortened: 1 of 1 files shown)


Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -28,6 +28,22 @@ final class ChatViewController: SLKTextViewController {
...
+            if showButtonScrollToBottom {
+                buttonScrollToBottomMarginConstraint?.constant = -64
+            } else {
+                buttonScrollToBottomMarginConstraint?.constant = 50
+            }
+
+            if showButtonScrollToBottom != oldValue {
+                UIView.animate(withDuration: 0.5) {
+                    self.view.layoutIfNeeded()
+                }
```
rafaelks: This is working great, the position is now being calculated correctly, but this line is causing a very weird animation. Maybe if you change from `self.view` to `self.buttonScrollToBottom` this won't happen.

PR #763, opened 2017-10-09
[FIX] Video receiving wrong extension when uploading

### LINKED ISSUE #754: [BUG] When uploading a video from the app, the extension is getting ".jpeg"
- Your Rocket.Chat app version: 1.5.0

cardoso:
@RocketChat/ios

Also changed the encoding to mpeg4, so it's now possible to view the videos in other clients (changing the extension wasn't enough)

Closes #754

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
@@ -68,7 +68,7 @@ extension ChatViewController: UIImagePickerControllerDelegate, UINavigationContr
 
     // MARK: UIImagePickerControllerDelegate
     func imagePickerController(_ picker: UIImagePickerController, didFinishPickingMediaWithInfo info: [String : Any]) {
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
```diff
@@ -110,8 +110,8 @@ extension ChatViewController: UIImagePickerControllerDelegate, UINavigationContr
 
                 file = UploadHelper.file(
                     for: videoData as Data,
-                    name: filename,
-                    mimeType: UploadHelper.mimeTypeFor(videoURL)
+                    name: "\(filename).mp4",
```
rafaelks: It looks like it's adding the extension even if filename already have it. Can we check that and only add the `.mp4` extension when it doesn't have it? You can try by simply uploading a video from your iPhone that you recorded your screen (using latest iOS 11). Also, looks like if the extension is `.MOV` it will be `.MOV.mp4`.

PR #779, opened 2017-10-13
[FIX] Sometimes, the API URL isn't the current server's

macabeus:
@RocketChat/ios

Cleses #783

The solution that I made is the has least impact in the code, but, I don't know if is the best solution.

Currently, is save on the Realm only the websocket URL of the server - it's bad, because we now have two URL: one for API, and other for the websocket.
I think that the one better solution is save at Realm only the basic URL of the server (`open.rocket.chat`, for example), and then create two computed property at `Auth`: one called `serverApiURL` and other called `serverSocketURL`.
Currently, is confuse the property `Auth.serverURL` and `ServerPersistKeys.serverURL`. It's about socket url, ou api url?

What I proposed will have a big impact on the code. I make this, or it's not necessary?


cardoso:
This makes sense and I will be reviewing soon. Thanks @macabeus.

Are you aware of any issue this fixes?

macabeus:
I created the issue that this pullrequest fix: #783

rafaelks:
Makes sense @macabeus! I just think we need to use the URL and URLComponents  classes to deal with URL changes. We need to make sure that we're quite flexible with this. For example: instead of using `replace https per wss` you can change the `url.protocol`, got it?

Also, we need to make sure we're always keeping the URL's path, port and other things to maintain compatibility with all kinds of URLs people come with.

cardoso:
Thanks again for this contribution @macabeus

You can base your test in this test we already have:

```swift
class ConnectServerViewControllerSpec: XCTestCase {
    func testNormalizeInputURL() {
        let vc = ConnectServerViewController()

        XCTAssertEqual(vc.normalizeInputURL("open.rocket.chat"), "https://open.rocket.chat", "will add https")
        XCTAssertEqual(vc.normalizeInputURL("https://open.rocket.chat"), "https://open.rocket.chat", "will return correct url")
        XCTAssertEqual(vc.normalizeInputURL("http://open.rocket.chat"), "https://open.rocket.chat", "will force https scheme")
        XCTAssertEqual(vc.normalizeInputURL("http://open.rocket.chat/path"), "https://open.rocket.chat/path", "will keep path")
        XCTAssertEqual(vc.normalizeInputURL("http://open.rocket.chat?query=test"), "https://open.rocket.chat?query=test", "will keep query")
        XCTAssertEqual(vc.normalizeInputURL("http://open.rocket.chat/path?query=test"), "https://open.rocket.chat/path?query=test", "will keep path & query")
        XCTAssertNil(vc.normalizeInputURL("https://"), "will return nil when hostless")
        XCTAssertNil(vc.normalizeInputURL(""), "will return nil when empty")
    }
}
```

cardoso:
@macabeus, since this is an important PR and we're sending the new version today, I've made the improvements myself.

Thanks again!!

PR #788, opened 2017-10-15
[FIX] App crashing when going back from adding new team

### LINKED ISSUE #780: [BUG] The app crash when gives up to add a new team
- Your Rocket.Chat app version: 1.6.0 (104) (develop branch)
- Your Rocket.Chat server version: 0.58.2
- Device (or Simulator) you're running with: Simulator iPhone SE and physical device iPhone SE

cardoso:
@RocketChat/ios

Closes #780

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -69,6 +69,8 @@ final class ChatViewController: SLKTextViewController {
     var messages: [Message] = []
     var subscription: Subscription! {
         didSet {
+            if subscription.isInvalidated { return }
+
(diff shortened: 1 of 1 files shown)


Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -69,6 +69,8 @@ final class ChatViewController: SLKTextViewController {
     var messages: [Message] = []
     var subscription: Subscription! {
         didSet {
+            if subscription.isInvalidated { return }
```
rafaelks: What do you think about a `guard` here instead?

rafaelks:
I'll merge it, but we should know why it's using an invalidated object also.

PR #805, opened 2017-10-18
[NEW] Add feature to open direct message by clicking on someone's name

### LINKED ISSUE #803: [NEW] Direct message by clicking on someone's name
Allow user's to click on someone's name to bring them into a direct message conversation.

macabeus:
@RocketChat/ios

Closes #803

I made exactly what was requested: open direct message by clicking on someone's name.
But, I don't know if it's the ideal. Maybe is better open a user's profile, and, in the user's profile, has a button to open direct message?

And, in `ChatMessageCell` I think that now has ambiguity in name of variables and functions.
Previously, we have only one long press gesture. And, now, we have two gestures. Then, the names like `longPressGesture` and `tapGesture` aren't expressive enough to me. Maybe is better rename to `longPressMessageGesture` and `tapUsernameGesture`?

**To fix (rafaelks)**

- [x] Rename the events/gestures to something more "clear"
- [x] Refactor in order to remove Realm dependency in `ChatControllerMessageCellProtocol`


Rocket.Chat/Controllers/Chat/ChatControllerMessageCellProtocol.swift
```diff
@@ -19,6 +20,33 @@ extension ChatViewController: ChatMessageCellProtocol {
         }
     }
 
+    func handleTapMessageCell(_ message: Message, view: UIView, recognizer: UIGestureRecognizer) {
+        guard let username = message.user?.name else { return }
+
+        func openDirectMessage() -> Bool {
+            guard let directMessageRoom = Realm.shared?.objects(Subscription.self).filter("name == '\(username)' && privateType == 'd'").first else { return false }
```
rafaelks: Let's remove this Realm dependency. You can add this as a method in `Subscription` model for now.

Rocket.Chat/Controllers/Chat/ChatControllerMessageCellProtocol.swift
```diff
@@ -10,6 +10,7 @@ import UIKit
 import SafariServices
 import MobilePlayer
 import SimpleImageViewer
+import RealmSwift
```
rafaelks: Remove this line, since we'll remove the Realm dependency here.

macabeus:
Ok. I removed Realm dependency in `ChatControllerMessageCellProtocol.swift` creating a new static method in `Subscription`. This method also can be used in #743

Rocket.Chat/Models/Subscription.swift
```diff
@@ -159,4 +159,18 @@ extension Subscription {
         return object
     }
 
+    static func find(name: String, subscriptionType: [SubscriptionType], realm: Realm) -> Subscription? {
+        let predicate = NSPredicate(
+            format: "name == %@ && privateType IN %@",
+            name, subscriptionType.map { $0.rawValue }
+        )
+
+        return Realm.shared?.objects(Subscription.self).filter(predicate).first
```
rafaelks: Why two methods if you're using Realm.shared? in both?

rafaelks:
Made one small comment again, and have a question: what do you think about adding this to the `username` label inside a message text? For example, if I send `@macabeus Hello` and tap in `@macabeus`? Maybe this can be added in a future PR if you think it's a different thing.

cardoso:
@rafaelks that will involve a different part of the code and different UIKit APIs. It's best to make it in a separate PR. :)

I opened #818 and #819

Rocket.Chat/Models/Subscription.swift
macabeus: Oops, I was sleepy when I was programming. Sorry 😶
I create a method that pass `Realm.shared` as default, to don't need import Realm in `ChatControllerMessageCellProtocol.swift`

macabeus:
@rafaelks I agree with @cardoso 
I think that is better made it's in other pull request.

Rocket.Chat/Models/Subscription.swift
```diff
@@ -159,4 +159,18 @@ extension Subscription {
...
     }
 
+    static func find(name: String, subscriptionType: [SubscriptionType], realm: Realm) -> Subscription? {
+        let predicate = NSPredicate(
+            format: "name == %@ && privateType IN %@",
+            name, subscriptionType.map { $0.rawValue }
+        )
+
+        return realm.objects(Subscription.self).filter(predicate).first
+    }
```
rafaelks: If we're only using the method that's using `Realm.shared`, why not only keep that one?

macabeus: I made it in this way because I wanted to keep the pattern with other method with same name: `find(rid: String, realm: Realm)`.
This method is similar with the method that I was writing, and have `realm` as a parameter. Then, to my new method follow it's, I chose to wrote a function that also receive `realm` parameter.

Maybe it is not necessary  and I keep only one method?

rafaelks: Yes, let's only keep what's being used. Thanks!

Rocket.Chat/Controllers/Chat/ChatControllerMessageCellProtocol.swift
```diff
@@ -19,6 +19,34 @@ extension ChatViewController: ChatMessageCellProtocol {
         }
     }
 
+    func handleUsernameTapMessageCell(_ message: Message, view: UIView, recognizer: UIGestureRecognizer) {
+        guard let username = message.user?.name else { return }
```
rafaelks: If you need the user's username, you need to use the property `username` instead of `name`.

Rocket.Chat/Controllers/Chat/ChatControllerMessageCellProtocol.swift
```diff
@@ -19,6 +19,34 @@ extension ChatViewController: ChatMessageCellProtocol {
...
+            guard let directMessageRoom = Subscription.find(name: username, subscriptionType: [.directMessage]) else { return false }
+
+            let controller = ChatViewController.shared
+            controller?.subscription = directMessageRoom
+
+            return true
+        }
+
+        // Check if already have a direct message room with this user
+        if openDirectMessage() == true { return }
```
rafaelks: Please, use our style of `blocks` in here...

```swift
if openDirectMessage() {
    return
}
```

rafaelks:
@macabeus Looking good! I fixed a bug where I could tap in a user that I was already in a DM with. Also, added the handler to avatar as well, so if you tap avatar, will open a DM with the user. 👍

PR #823, opened 2017-10-21
[WIP][NEW] Open chat by tapping channel/mention reference in message

macabeus:
@RocketChat/ios

Closes #818 
Closes #819

- [x] Add gesture tap action when tap in channel reference
- [x] Add gesture tap action when tap in mention reference
- [ ] Create tap gesture only for valid channel/user name
- [ ] Open direct message when tap in mention reference
- [ ] Open channel when tap in reference of its

PR #834, opened 2017-10-22
[IMPROVEMENT] Rounded background for mentions/highlights

### LINKED ISSUE #633: [IMPROVEMENT] Improve @mentions styling
We are currently using NSAttributedString to achieve the current visuals which doesn't easily allow for padding and round corners by default.

| Currently (iOS) | Desired (Web) |
| - | - |
|  |  |

Where to look:
[NSAttributedStringExtensions.swift: 69](https://github.com/RocketChat/Rocket.Chat.iOS/blob/master/Rocket.Chat/Extensions/NSAttributedStringExtensions.swift#L69-L80)

artrmz:
@RocketChat/ios

Closes #633

mentions-highlight

**To fix**
- [x] Add rounded background for mentions and highlights

- [x] Clean code

- [x] Adjust tests to background changes

- [x] Increase line spacing when there is mention/highlight in message

- [x] Fix background frame positioning for long messages

- [x] Align text to username's leading constraint

- [x] Fix content drawing after rotation from landscape to portrait

- [x] Fix text selection

- [x] lower the background for mentions to make it look more separated from each other


macabeus:
I liked your class `HighlightTextView`. It's useful for the pull request #823 in order to make `ChatMessageCell` more readable.
[I merged your branch with my branch](https://github.com/macabeus/Rocket.Chat.iOS/commit/977ee81fa056a5cf799a6ec2f0f4bf76de4d5772), to also use this class, then, when your pull request is merged, our features will be working together.
Thank you!

artrmz:
Added few changes. Also added increasing line spacing for whole message if it contains any highlight to make it look better (otherwise background was overlapping each other in some cases). Maybe we should add 10% of font size line spacing for all messages on the other hand?

Rocket.Chat/Views/Chat/HighlightTextView.swift
```diff
@@ -0,0 +1,80 @@
...
+        for index in 0..<lines.count {
+            // swiftlint:disable force_cast
+            let line = lines[index] as! CTLine
+            // swiftlint:enable force_cast
+
+            let glyphRuns = CTLineGetGlyphRuns(line) as NSArray
+
+            for i in 0..<glyphRuns.count {
+                // swiftlint:disable force_cast
+                let run = glyphRuns[i] as! CTRun
```
rafaelks: Let's not disable any `force_cast`

artrmz: 👍

artrmz:
@rafaelks done as you suggested, although I am not sure if we can merge it until #878 is fixed anyway since it's kinda blocking my tests. Wanted to test it for very long msg with lots of mentions but right now can't do it (same issue on App Store version)

rafaelks:
@artrmz Looks great, thank you! Yes, let's fix this bug first before merging your PR.

artrmz:
Fixed #878 although gotta fix positioning a little bit more since it's a bit off for very long messages  (it starts to get off by few pixels after like 20th line of text)

artrmz:
Seems like it's all fixed now ;) Waiting for reviews. Let me know if should I resolve conflicts or you will do that while merging to develop.

@rafaelks @cardoso

rafaelks:
@artrmz Thank you for your awesome work! Just a minor thing... look at the screenshot, you see that the `mention` isn't aligned to the `username` on the top? Can you fix this, please?

img_0150

artrmz:
Seems like easy thing to fix, will do that today or tomorrow :)

artrmz:
@rafaelks Fixed the aligmnent although found another issue ^^ It draws incorrectly (not always) after rotation.

artrmz:
Found a workaround for rotation issue although there is another issue with UITextView (probably on maste/develop branch as well) causing to resize incorrectly (wrong height) and cutting a text (last word of message) in some rare situations. Gotta investigate it further before I finish this one. I feel like writing my own UITextView (subclassing from UIView instead of UITextView) would be easier and give the better results in this situation 😱

cardoso:
@artrmz 

Text cutting can be fixed by tweaking this function in ChatMessageCell.swift:

```swift
    static func cellMediaHeightFor(message: Message, sequential: Bool = true) -> CGFloat {
        let fullWidth = UIScreen.main.bounds.size.width
        let attributedString = MessageTextCacheManager.shared.message(for: message)
        let height = attributedString?.heightForView(withWidth: fullWidth - 62)

        var total = (height ?? 0) + (sequential ? 8 : 29)
```

Tweaking some values there like adding 1 or 2 pixels to the height should fix it.

artrmz:
@cardoso Ah, thanks for explaining. That makes it a little bit more complicated. Although I think we shouldn't use some constants (62 a example here) like that. Should have taken those from view's constraints and subtract additional 16 for UITextView's scrollView insets if I am correct. That way we won't have to touch that method in case someones edits storyboard.

rafaelks:
@artrmz Yes! We may change the way we handle the size of the cell by just using AutoLayout, I'm not opposite to this idea, but will require a few changes in the code.

rafaelks:
@artrmz We changed the code a little bit, to process markdown in background & update on main-thread only. Do you mind fixing the conflicts? Maybe we could add this PR to `1.7.3` milestone. :-)

artrmz:
@rafaelks Yes, I'll solve all conflicts although it requires a bit more work until it's ready for merging.

rafaelks:
@artrmz This PR is looking really good, are we missing something to merge it? Really wanted to merge this new style. 👍

artrmz:
@rafaelks It breaks few things sadly and I did not have time to take care of it yet. For example text selection does not work.

cardoso:
@artrmz looks like we don't need selection anymore 😃 #1001

artrmz:
@cardoso Great :) Will take a look at this PR again since seems like it broke after all those changes on develop, probably something with mentions handling in the background.

artrmz:
I've refactored it completely. Should be ready to test "soon" :)

rafaelks:
@artrmz Thank you!!! 🎉

rafaelks:
@artrmz Thank you for your PR! I was testing here, and looks like the spacing between messages is too big. Do you also experience that?

simulator screen shot - iphone se - 2017-11-30 at 09 49 42

artrmz:
@rafaelks Will check this out. Thanks!

artrmz:
@rafaelks Should be fixed now
zrzut ekranu 2017-11-30 o 14 28 25

artrmz:
Restored isSelectable as #1027 needed it.

rafaelks:
@artrmz This is working great to me know... may I ask you just one more thing? The left-edge and right-edge on a mention background, can it be a little bigger? Because the spacing right now between the background color and the font is almost 0px.

Example here:

chat

artrmz:
@rafaelks The easiest way to achieve that would be to add nonbreaking spaces at start/end of the mention. Dunno if we want to do such thing? Or can try to add few pixels on each side, will give it a try tomorrow.

artrmz:
@rafaelks That's best I could achieve at the moment.

zrzut ekranu 2017-12-06 o 09 02 05
zrzut ekranu 2017-12-06 o 09 02 02

I think it looks better without those changes ;) It will either clip to bounds and be cut if mention is the first word in the line OR can add padding to whole text but it won't be aligned with the username. It will be really hard to detect if first word in line is a mention and move the whole line.

rafaelks:
@artrmz Currently way looks better, but only because it's cutting on the left... let's merge this way now and try to improve it over the time! 👍. Thanks a lot for looking into it.

PR #836, opened 2017-10-23
[FIX] A transparent square in the middle of the screen captures the gestures

### LINKED ISSUE #835: [BUG] A transparent square in the middle of the screen captures the gestures
- Your Rocket.Chat app version: develop branch - 1.6.0 (108)
- Device (or Simulator) you're running with: Simulator iPhone SE

image

I really have no idea, but, for some reason, Chat Screen has a transparente square.
The problem os this square is that it capture the gesture in Chat Screen, making problems in features like #823. For example, in the photo I can't tap in #foo and #general references.

macabeus:
@RocketChat/ios

Closes #835


rafaelks:
Thank you for finding that @macabeus!

PR #849, opened 2017-10-24
[IMPROVEMENT] Hide ssl requirement for custom servers

sowenjub:
@RocketChat/ios

Hey guys, I feel like displaying the ssl requirement label below the server url textfield doesn't make sense when user can't change it. This PR simply hides it when there's a RC_SERVER_URL set.


rafaelks:
@sowenjub Thank you for your PR, but I think we should not only change this part if we want to have a custom UI for when the server is **static**. What about not even showing the server screen and going directly to the authentication? Sounds better to me... what do you think?

sowenjub:
Hey @rafaelks, it does sound better, but it requires more discussion/work, hence this solution :)
ConnectServerViewController#connect fetch settings if I understand correctly, so we need a "loading" state before a user can enter their credentials, which might as well be that screen.

If you'd prefer to skip the screen entirely, here are my questions:

-  I'd probably hide the inputs and add a loading indicator. Does that sound fine?
- we'd have something similar to ConnectServerViewController#connect in AuthViewController but without the "Check if server already exists and connect to that instead" part of ConnectServerViewController#connect, right?

Thanks

rafaelks:
@sowenjub I think you're right about hiding things... you can actually use the same controller `ConnectServerViewController.swift` but make a few adjustments:

1. Yes, show the loader that server is connecting;
2. Hide the SSL label;
3. Hide the textField that contains the URL;
4. Add a label that will show the URL that user is connecting, but only the **host** part of it (e.g for *https://open.rocket.chat* it would only show *open.rocket.chat*)

What do you think?

sowenjub:
@rafaelks Sounds good to me, we might want to add a localized "Connecting to" label above the host label as well

rafaelks:
@sowenjub Yes, that would be 💯

sowenjub:
@rafaelks Thinking of it, I wonder if we shouldn't do something closer to the existing interface to avoid adding complexity to the storyboard controller.

So for static servers, we can simply:

- replace "It is required that your server supports SSL" with "Connection in progress. Hold tight."
- set textFieldServerURL.isEnabled to false
- display the host part in textFieldServerURL

As an added bonus, "It is required that your server supports SSL" would be moved to Localizable.strings 😃 

Let me know what you think.

sowenjub:
Also, can my PR get rid of [this line](https://github.com/RocketChat/Rocket.Chat.iOS/blob/1aba0fcdf2993e108925ca8b39baec79fa829bf0/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift#L137) ?

It doesn't seem to have any real use, but for our case, it means that we can't simply display `open.rocket.chat` without another check, since it prefixes it with "https://"

rafaelks:
@sowenjub Sounds great to me! Let's go with your suggestion! 👍

PR #850, opened 2017-10-24
[NEW] Disable registration

sowenjub:
@RocketChat/ios

This PR adds a new RC_DISABLE_REGISTRATION option to the AppManager, which can be useful for private servers where self-registration isn't allowed.
It simply hides all views related to the creation of a new account if this is set to true. If the option is absent or if it's set to false, it's just business as usual.


rafaelks:
@sowenjub We appreciate your PR, but this is something already implemented in our servers as a setting `Accounts_RegistrationForm`. Can you please, update your implementation to follow the `Accounts_RegistrationForm` value? The options are: `Public`, `Disabled` and `Secret URL`. It's OK if you only want to implement `Public` and `Disabled` for now.

sowenjub:
@rafaelks 👍

sowenjub:
@rafaelks Updated to use the Accounts_RegistrationForm. I assumed people using a secret url wouldn't want users who simply downloaded the app to be able to register.

rafaelks:
Looking good @sowenjub! Just need to resolve the conflicts now, I suppose. Thank you!

sowenjub:
@rafaelks [https://github.com/RocketChat/Rocket.Chat.iOS/pull/850/commits/a5646ba7bae1d433c0c9bd431bb8a34a3abdbc1d](this merge) should fix it, but it stills displays the same conflicts, I'm a bit puzzled as to what to do next

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -65,6 +67,7 @@ final class ConnectServerViewController: BaseViewController {
         DatabaseManager.cleanInvalidDatabases()
 
         if let applicationServerURL = AppManager.applicationServerURL {
+            labelSSLRequired.isHidden = true
```
rafaelks: This should not be in this PR, right?

Rocket.Chat/Managers/AppManager.swift
```diff
@@ -35,5 +35,4 @@ struct AppManager {
     static var supportsMultiServer: Bool {
         return applicationServerURL == nil
     }
-
```
rafaelks: 🤔

Rocket.Chat/Models/AuthSettings.swift
```diff
@@ -30,6 +30,17 @@ final class AuthSettings: BaseModel {
     @objc dynamic var isGoogleAuthenticationEnabled = false
     @objc dynamic var isLDAPAuthenticationEnabled = false
 
+    // Registration
+    enum RegistrationFormAccess: String {
+        case isPublic = "Public", isDisabled = "Disabled", isSecretURL = "Secret URL"
```
rafaelks: 1. Break each case per line, please
2. Move the enum to outside of the class, please (top)

Rocket.Chat/Models/AuthSettings.swift
```diff
@@ -30,6 +30,17 @@ final class AuthSettings: BaseModel {
...
+    }
+    @objc dynamic var rawRegistrationForm: String?
+    var registrationForm: RegistrationFormAccess {
+        guard let rawValue = rawRegistrationForm,
+            let value = RegistrationFormAccess(rawValue: rawValue) else { return .isPublic }
```
rafaelks: Could you change for this?

```swift
guard let rawValue = rawRegistrationForm,
      let value = RegistrationFormAccess(rawValue: rawValue)
else {
     return .isPublic
}

File: Rocket.Chat/Controllers/Auth/AuthViewController.swift [hunk 5]
```diff
@@ -47,6 +47,8 @@ final class AuthViewController: BaseViewController {
 
     @IBOutlet weak var activityIndicator: UIActivityIndicatorView!
 
+    @IBOutlet var accountCreationFields: [UIView]!
```
rafaelks: What's `accountCreationFields`? Shouldn't it be just the `Register button`?

sowenjub: Changed it to just the button

Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
sowenjub: Right

Rocket.Chat/Managers/AppManager.swift
sowenjub: A space disappeared when I switched from using a custom Info.plist to server values. I put it back 😛

Rocket.Chat/Models/AuthSettings.swift
sowenjub: Alright.
Not sure about making that enum global but that's your call.

Rocket.Chat/Models/AuthSettings.swift
sowenjub: Ok for aligning vertically but are you sure about the else on a new line? That's the only change I didn't make 😈

rafaelks: @sowenjub Yes, please! 😂 That's how we're doing in other places of the app for guards with multiple lines: https://github.com/RocketChat/Rocket.Chat.iOS/blob/1aba0fcdf2993e108925ca8b39baec79fa829bf0/Rocket.Chat/OAuthManager.swift#L50-L61.

sowenjub: Alright, alright, pushed

rafaelks:
Thanks!!!

rafaelks:
This is looking good now @sowenjub, just added a few tests, will wait for CI and merge it, thanks! 👍

sowenjub:
Awesome, thanks for adding the tests @rafaelks

PR #853, opened 2017-10-24
[Fix] Use guard instead of if statement.

JourKung:
I used guard instead of if for unwrapping optional values.


cardoso:
I'm sorry, but this doesn't fix anything and, as you can see, actually breaks our tests.

This will have to be closed for now as we already have a large amount of PRs to review.

If you still feel inclined to contribute, I suggest you look into our list of issues. Thanks.

PR #857, opened 2017-10-25
[FIX] Attachments with special characters not opening

### LINKED ISSUE #851: [BUG] Filenames with umlauts cause broken upload
@emign commented on [Tue Oct 24 2017](https://github.com/RocketChat/Rocket.Chat.Electron/issues/579)

On rocketchat+ on iOS when you upload a file (eg pdf) with an umlaut the file seems to upload correctly and is shown in the message stream. But clicking it does not open it but only shifts the text up and down a bit 

---

@gdelavald commented on [Tue Oct 24 2017](https://github.com/RocketChat/Rocket.Chat.Electron/issues/579#issuecomment-338983212)

@emign Sorry, is this mobile or on a Mac computer?

---

@emign commented on [Tue Oct 24 2017](https://github.com/RocketChat/Rocket.Chat.Electron/issues/579#issuecomment-338984487)

It is on iOS mobile (iPad and iPhone) with version 1.6.0. the same (umlaut-titled file) can be downloaded via a browser on the device just fine btw.

cardoso:
@RocketChat/ios

Closes #851

### DIFF
--- a/Rocket.Chat/Models/Attachment.swift
+++ b/Rocket.Chat/Models/Attachment.swift
@@ -61,13 +61,20 @@ class Attachment: BaseModel {
 extension Attachment {
 
     fileprivate static func fullURLWith(_ path: String?) -> URL? {
-        guard let path = path?.replacingOccurrences(of: "//", with: "/") else { return nil }
-        guard let auth = AuthManager.isAuthenticated() else { return nil }
-        guard let userId = auth.userId else { return nil }
-        guard let token = auth.token else { return nil }
-        guard let baseURL = auth.baseURL() else { return nil }
-        let pathNoSpaces = path.replacingOccurrences(of: " ", with: "%20")
-        let urlString = "\(baseURL)\(pathNoSpaces)?rc_uid=\(userId)&rc_token=\(token)"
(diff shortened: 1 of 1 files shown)


rafaelks:
@cardoso Super!!! Can we also apply the same changes to the imageURL and videoURL? I think the problem also happens for it.

cardoso:
@rafaelks 

```swift
    func fullFileURL() -> URL? {
        return Attachment.fullURLWith(titleLink)
    }

    func fullVideoURL() -> URL? {
        return Attachment.fullURLWith(videoURL)
    }

    func fullImageURL() -> URL? {
        guard let imageURL = imageURL else { return nil }
        if imageURL.contains("http://") || imageURL.contains("https://") {
            return URL(string: imageURL)
        }

        return Attachment.fullURLWith(imageURL)
    }

    func fullAudioURL() -> URL? {
        return Attachment.fullURLWith(audioURL)
    }
```

They all call the same function, which is the one I adjusted. Or is it something else? 🤔

rafaelks:
Right!!! In the past it was three separated functions, but we already optimized that! 👍

PR #861, opened 2017-10-25
[FIX] Read-only rooms not being handled correctly

### LINKED ISSUE #848: [BUG] Read-only rooms not being properly handled
Bug preview:

We should block the user from sending messages if it's a read only room **and** he has no permission.

The text field should be disabled and the user should be made aware that this is a read only room.

cardoso:
@RocketChat/ios

|Preview|
|-|
||

Closes #848

### DIFF
--- a/Rocket.Chat/Models/Mapping/SubscriptionModelMapping.swift
+++ b/Rocket.Chat/Models/Mapping/SubscriptionModelMapping.swift
@@ -23,6 +23,7 @@ extension Subscription: ModelMappeable {
         self.open = values["open"].bool ?? false
         self.alert = values["alert"].bool ?? false
         self.favorite = values["f"].bool ?? false
+        self.readOnly = values["ro"].bool ?? false
 
         if let typeString = values["t"].string {
             self.type = SubscriptionType(rawValue: typeString) ?? .channel
--- a/Rocket.Chat/Models/Mapping/SubscriptionModelMapping.swift
+++ b/Rocket.Chat/Models/Mapping/SubscriptionModelMapping.swift
@@ -49,5 +49,9 @@ extension Subscription: ModelMappeable {
         if let readOnly = values["ro"].bool {
             self.roomReadOnly = readOnly
         }
+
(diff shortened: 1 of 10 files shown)


Rocket.Chat/Models/Mapping/SubscriptionModelMapping.swift
```diff
@@ -41,4 +41,17 @@ extension Subscription: ModelMappeable {
             self.lastSeen = Date.dateFromInterval(lastSeen)
         }
     }
+
+    func mapRoom(_ values: JSON) {
```
rafaelks: Let's add some tests to this method? :-)

Rocket.Chat/Models/Subscription.swift
```diff
@@ -45,11 +45,29 @@ class Subscription: BaseModel {
...
+    @objc dynamic var roomOwnerId: String?
+    var roomOwner: User? {
+        guard
+            let realm = Realm.shared,
+            let roomOwnerId = roomOwnerId
+        else {
+            return nil
+        }
+
+        return realm.objects(User.self).filter("identifier = '\(roomOwnerId)'").first
```
rafaelks: Let's also add some tests to this property?

Rocket.Chat/Models/Subscription.swift
```diff
@@ -45,12 +45,18 @@ class Subscription: BaseModel {
...
+        guard let roomOwnerId = roomOwnerId else { return nil }
+        return User.find(withIdentifier: roomOwnerId)
+    }
 
     @objc dynamic var otherUserId: String?
     var directMessageUser: User? {
-        guard let realm = Realm.shared else { return nil }
         guard let otherUserId = otherUserId else { return nil }
-        return realm.objects(User.self).filter("identifier = '\(otherUserId)'").first
+        return User.find(withIdentifier: otherUserId)
```
rafaelks: 👏

rafaelks:
SUPER!!!

PR #876, opened 2017-10-27
[i18n] Add Czech localization

flashbackcz:
Hi,
I completely translated english strings to czech.

It would be nice to see the translation in the next version of the application. (:


rafaelks:
@flashbackcz Thank you for your PR! But looks like it wasn't working, but your file was written correctly. So I made a few adjustments to make sure it works. 👍

PR #886, opened 2017-10-30
[FIX] Members list title is wrong for other languages

### LINKED ISSUE #881: [BUG] Members list title resource is wrong for other supported languages
The title is wrong while data is being loaded/downloaded. Title should not be visible at the very start until data is downloaded or should be replaced for proper resource before controller appears and data is downloaded.

simulator screen shot - iphone x - 2017-10-29 at 14 41 49

cardoso:
@RocketChat/ios

Closes #881

### DIFF
--- a/Rocket.Chat/Controllers/Chat/MembersListViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/MembersListViewController.swift
@@ -17,9 +17,7 @@ class MembersListViewData {
     var showing: Int = 0
     var total: Int = 0
 
-    var title: String {
-        return String(format: localized("chat.members.list.title"), total)
-    }
+    var title: String = localized("chat.members.list.title")
 
     var isShowingAllMembers: Bool {
         return showing >= total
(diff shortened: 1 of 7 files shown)


Rocket.Chat/Controllers/Chat/MembersListViewController.swift
```diff
@@ -110,6 +110,8 @@ extension MembersListViewController {
         if let cell = membersTableView.dequeueReusableCell(withIdentifier: LoaderTableViewCell.identifier) as? LoaderTableViewCell {
             self.loaderCell = cell
         }
+
+        title = data.title
```
rafaelks: Isn't the title look like `Members (0)` this way?

cardoso:
I'm not sure that's what we want, @dennispost. @rafaelks?

dennispost:
@cardoso I'm down with you. There should be always a title. I just wanna say that #881 means a different way. 

If we want to continue to work on this screen I think we need something like this [DBEmptyState](https://github.com/dbsystel/DBEmptyState).

rafaelks:
@cardoso Yes, your PR looks good! I also liked @dennispost suggestion to improve our empty state on lists of messages, members, files, etc. But I don't think we need a library to do that right now, we could implement that by ourselves... what do you guys think?

artrmz:
@rafaelks, agree, should be in other idzie. I would love to work on that once I have a little bit of free time.

@cardoso, thank you for your PR.

PR #898, opened 2017-10-31
[FIX] User being able to send messages when muted

### LINKED ISSUE #862: [BUG] Muted state not being handled correctly
Currently, the sent message just becomes blank and the message from the BOT is not displayed

This is what happens in the web version:

Maybe we should try to handle this differently. 
The same way we'll handle read-only channels: #861

What do you think @rafaelks?

cardoso:
@RocketChat/ios

Closes  #862

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -63,6 +63,8 @@ final class ChatViewController: SLKTextViewController {
     var isRequestingHistory = false
     var isAppendingMessages = false
 
+    var subscriptionToken: NotificationToken?
+
     let socketHandlerToken = String.random(5)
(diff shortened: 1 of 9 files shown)


Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -71,6 +73,20 @@ final class ChatViewController: SLKTextViewController {
         didSet {
             guard !subscription.isInvalidated else { return }
 
+            subscriptionToken?.invalidate()
```
rafaelks: Needs to invalidate the token here too: https://github.com/RocketChat/Rocket.Chat.iOS/blob/9ed56d60e288823a96f920bf68754f49361d0e10/Rocket.Chat/Controllers/Chat/MainChatViewController.swift#L81-L96

PR #906, opened 2017-11-01
[FIX] Loading forever if a socket error happened in connection

### LINKED ISSUE #890: [IMPROVEMENT] Loading forever with no user feedback when adding a new server
- Your Rocket.Chat app version: 1.7.0 (109)
- Your Rocket.Chat server version: ?
- Device (or Simulator) you're running with: iPhone SE Simulator

I'm getting stuck in this loading when trying to connect to some servers.

screen shot 2017-10-30 at 20 14 23

The problem is not that there is a connection error but it seems that the error is not being handled properly. I could see that always when the following web socket error appears on the console the loading stuck and there is no feedback for the user:

screen shot 2017-10-30 at 20 14 26

cardoso:
@RocketChat/ios

Closes #890

### DIFF
--- a/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
@@ -180,7 +180,8 @@ final class ConnectServerViewController: BaseViewController {
             SocketManager.connect(socketURL) { (_, connected) in
                 if !connected {
                     self?.stopConnecting()
-                    self?.alert(title: "Connection Error", message: "Could not connect to this server")
+                    self?.alert(title: localized("alert.connection.socket_error.title"),
+                                message: localized("alert.connection.socket_error.message"))
                     return
                 }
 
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Controllers/Auth/ConnectServerViewController.swift
```diff
@@ -170,28 +170,32 @@ final class ConnectServerViewController: BaseViewController {
...
             }
 
-            let index = DatabaseManager.createNewDatabaseInstance(serverURL: socketURL.absoluteString)
-            DatabaseManager.changeDatabaseInstance(index: index)
             SocketManager.connect(socketURL) { (_, connected) in
+                if !connected {
+                    self?.stopConnecting()
+                    self?.alert(title: localized("alert.connection.socket_error.title"),
+                                message: localized("alert.connection.socket_error.message"))
+                    return
```
rafaelks: 👍

PR #934, opened 2017-11-07
[NEW] Video recording

### LINKED ISSUE #720: [NEW] Enable recording and sending videos directly
Version: 1.4.1 (99) 

At the moment it is not possible to record and send a video directly. We have to go through the default camera app to record a video to the gallery and can then send it through the rocket.chat native app.

It would be super cool if we can record and send it directly like it works for photos.

Thanks!

cardoso:
@RocketChat/ios

Closes #720

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
@@ -19,6 +19,14 @@ extension ChatViewController: UIImagePickerControllerDelegate, UINavigationContr
             alert.addAction(UIAlertAction(title: localized("chat.upload.take_photo"), style: .default, handler: { (_) in
                 self.openCamera()
             }))
+
+            alert.addAction(UIAlertAction(title: localized("chat.upload.shoot_video"), style: .default, handler: { (_) in
+                AVAudioSession.sharedInstance().requestRecordPermission { granted in
+                    if granted {
+                        self.openCamera(video: true)
+                    }
+                }
+            }))
         }
 
         alert.addAction(UIAlertAction(title: localized("chat.upload.choose_from_library"), style: .default, handler: { (_) in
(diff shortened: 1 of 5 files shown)


Rocket.Chat/pt-BR.lproj/Localizable.strings
```diff
@@ -102,6 +102,7 @@
 "chat.upload.uploading_video" = "Enviando video";
 "chat.upload.uploading_file" = "Enviando \"%@\"";
 "chat.upload.take_photo" = "Tirar uma foto";
+"chat.upload.shoot_video" = "Filmar um video";
```
rafaelks: "Gravar um vídeo"?

Rocket.Chat/Controllers/Chat/ChatControllerUploader.swift
```diff
@@ -19,6 +19,14 @@ extension ChatViewController: UIImagePickerControllerDelegate, UINavigationContr
             alert.addAction(UIAlertAction(title: localized("chat.upload.take_photo"), style: .default, handler: { (_) in
                 self.openCamera()
             }))
+
+            alert.addAction(UIAlertAction(title: localized("chat.upload.shoot_video"), style: .default, handler: { (_) in
+                AVAudioSession.sharedInstance().requestRecordPermission { granted in
+                    if granted {
+                        self.openCamera(video: true)
+                    }
```
rafaelks: If not, let's show an error message?

cardoso: Looks like camera handles that for us, so I'm removing this part :)

Rocket.Chat/pt-BR.lproj/Localizable.strings
cardoso: 👍

PR #935, opened 2017-11-07
[IMPROVEMENT] Highlight performance

### LINKED ISSUE #931: [IMPROVEMENT] Highlight performance
We have very bad performance when parsing highlights.

cardoso:
@RocketChat/ios

Closes #931 

This solved lots of redundant calls we were having:
```swift
if obj.message?.text != message.text {
        MessageTextCacheManager.shared.update(for: message)
}
```
Sometimes some metadata is updated, but doesn't mean the text is :)

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatDataController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatDataController.swift
@@ -196,7 +196,10 @@ final class ChatDataController {
     func update(_ message: Message) -> Int {
         for (idx, obj) in data.enumerated()
             where obj.message?.identifier == message.identifier {
-                MessageTextCacheManager.shared.update(for: message)
+                if obj.message?.text != message.text {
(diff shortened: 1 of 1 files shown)


Rocket.Chat/Controllers/Chat/ChatDataController.swift
```diff
@@ -196,7 +196,10 @@ final class ChatDataController {
     func update(_ message: Message) -> Int {
         for (idx, obj) in data.enumerated()
             where obj.message?.identifier == message.identifier {
-                MessageTextCacheManager.shared.update(for: message)
+                if obj.message?.text != message.text {
```
rafaelks: 😱

rafaelks: Maybe add this to the `where` clause?

PR #946, opened 2017-11-08
[REFACTOR] quoteString and replyString + add tests

cardoso:
@RocketChat/ios

Made replyString and quoteString computed variables in Message instead of methods in ChatViewController and added tests for them.

Also added some test utilities


rafaelks:
Great! Thank you!!!

PR #959, opened 2017-11-14
[NEW] Handle errors & redirect into API requests

### LINKED ISSUE #937: [NEW] Handle HTTP redirects (301, 308) and update the server list accordingly
# Description

When server changes it's URL the app needs to ask the user if user wants to change it or not, showing the new URL.

So, the app needs to do the following things:

1. Before connecting to the web-socket, check `/api/info` always and see server is valid;
2. If response is `301` or `308` status code, app needs to check if there's a `Location` header in the 
3. If `Location` is a valid URL, needs to check if it's a valid Rocket.Chat server, by requesting `/api/info`. This process is already being done on the `ConnectViewController.swift` file;
4. If new server is valid, user needs to be prompted with an alert asking if user want's to change the old URL to the new one;
5. If user taps "Yes", the app needs to clear all the database, change server information to the new URL and keep the auth information;

rafaelks:
@RocketChat/ios

Closes #937

## Description

This PR is suppose to add a way to handle errors and redirect to all of our REST API calls. It will also add an alert when we request `/api/info` and server tries to redirect, alerting user that server's URL changed.

## Progress

- [x] Add error & statusCode to APIResult
- [x] Call completion when URLSession notifies a redirect
- [x] i18n
- [x] Unit tests


Rocket.Chat/Controllers/Chat/MembersListViewController.swift
```diff
@@ -38,7 +38,7 @@ class MembersListViewData {
 
         if let subscription = subscription {
             isLoadingMoreMembers = true
-            API.shared.fetch(SubscriptionMembersRequest(roomId: subscription.rid, type: subscription.type), options: .paginated(count: pageSize, offset: currentPage*pageSize)) { result in
+            API.shared.fetch(SubscriptionMembersRequest(roomId: subscription.rid, type: subscription.type), options: .paginated(count: pageSize, offset: currentPage*pageSize), { result in
```
cardoso: Can we do the same thing here :top:

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
```diff
@@ -53,8 +53,11 @@ class MessagesListViewData {
 
         if let subscription = subscription {
             isLoadingMoreMessages = true
-            API.shared.fetch(SubscriptionMessagesRequest(roomId: subscription.rid, type: subscription.type, query: query),
-                             options: .paginated(count: pageSize, offset: currentPage*pageSize)) { result in
+
+            let request = SubscriptionMessagesRequest(roomId: subscription.rid, type: subscription.type, query: query)
+            let options = APIRequestOptions.paginated(count: pageSize, offset: currentPage*pageSize)
+
+            API.shared.fetch(request, options: options, { result in
```
cardoso: That you did here 🔝 ?

Rocket.Chat/Managers/Requests/InfoRequestHandler.swift
```diff
@@ -0,0 +1,134 @@
...
+        if let location = response.allHeaderFields["Location"] as? String {
+            if let newURL = transformNewURL(location) {
+                handleRedirect(newURL)
+            }
+        }
+
+        completionHandler(nil)
+    }
+
+    func transformNewURL(_ newURL: String) -> URL? {
```
cardoso: We have URL.init(string:, scheme:) for this. Can we remove this and use that instead?

Rocket.ChatTests/Managers/DatabaseManagerSpec.swift
```diff
@@ -9,26 +9,40 @@
 import XCTest
 @testable import Rocket_Chat
 
+private let testServers = [[
```
cardoso: Is this global 😱 ? Should we move it to the DatabaseManager extension?

rafaelks: You actually can't add properties to the extensions... let's leave it there until we find a better solution?

Rocket.Chat/Controllers/MainViewController.swift
```diff
@@ -98,4 +76,82 @@ final class MainViewController: BaseViewController {
...
+        DatabaseManager.changeDatabaseInstance()
+        AuthManager.recoverAuthIfNeeded()
+
+        DispatchQueue.main.async {
+            if let auth = AuthManager.isAuthenticated() {
+                if let apiHost = auth.apiHost {
+                    self.infoRequestHandler.validate(with: apiHost)
+                }
+            }
+        }
```
cardoso:
```swift
DispatchQueue.main.async {
    guard let auth = AuthManager.isAuthenticated() else { return }
    guard let apiHost = auth.apiHost() else { return }
    self.infoRequestHandler.validate(with: apiHost)
 }
```

Rocket.Chat/Controllers/Chat/MembersListViewController.swift
rafaelks: Sure thing!

Rocket.ChatTests/Managers/DatabaseManagerSpec.swift
cardoso: Always forget that 👍

Rocket.Chat/Managers/Requests/InfoRequestHandler.swift
rafaelks: Done, good finding!

Rocket.Chat/Controllers/MainViewController.swift
rafaelks: Done!

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
rafaelks: Done! 👍

PR #963, opened 2017-11-14
[FEATURE] Invite user while creating channel

### LINKED ISSUE #902: [NEW] Add user to a channel you creating
- Your Rocket.Chat app version: 1.7.0 (110)
- Your Rocket.Chat server version: 0.59.3
- Device (or Simulator) you're running with: iPhone 6 Plus

At the moment we can create new channels from the iPhone App, but couldnt invite new users to the channel.

artrmz:
@RocketChat/ios

Closes #902 

zrzut ekranu 2017-11-20 o 12 28 20

- [x] show users that will be invited

- [x] fix cell height if there is no users invited yet

- [x] fetch users that fits to typed in string

- [x] show the list of fetched users

- [x] use API to invite users

- [x] add localizes

- [x] scroll table view if textField falls under keyboard

- [x] add users searching throttle to not spam network while fast typing

- [x] add tests


artrmz:
@rafaelks 

My query for `/api/v1/users.list` is `self.query = "query={ \"name\": { \"$regex\": \"\(name)\" } }"`
and lets say there is few people on server named like (I am not sure if the issue can be reproduced with this exact example, it's just random example since I can't give you names from my server):

- artrmz
- army
- adriano
- angelo
- adam
- brian
- paul

Let's say I want to search for "artrmz". When I query for 'a' it gives me all the names above, when I type "ar" it gives me correctly 2 first names but when I type "art" it gives me 0 results. Seems like the more characters I type the exact result sometimes won't be matched.

artrmz:
@rafaelks @cardoso Only tests left to do although I have used as a pod https://github.com/apasccon/SearchTextField which seems to be buggy sometimes so don't know, maybe I should write my own textField that works similar to this one?

rafaelks:
@artrmz Looking good! About the SearchTextField, I would say that’s much better if we could implement it, because it will probably be used in other parts of the app. If you’re feeling adventurous... go ahead, please! :-)

artrmz:
@rafaelks I have added tests (however Codecov seems showing incorrect coverage? how comes it loses so much percents 😬 - Xcode shows 32.19%)

I'll keep that SearchTextField pod for now. I do have plans to write my own implementation of such textField but since I am short on time it'll take few days (or even weeks). We can go with what we have now and improve it later on once I am done with my own text field.

cardoso:
@artrmz Codecov seems to be randomly jumping between 20~30% coverage on PRs. Let's ignore it :)

cardoso:
I'm not being able to create channels anymore 😵 Nothing happens when I press "create".

Develop is working fine.

rafaelks:
@cardoso Are you adding a name to the channel? I was having the same problem 🙈. It would be good if we disabled the button or alert the user with the error that's happening.

artrmz:
Will check this out when I am home. Was working fine last time I tried, maybe something changed in the meantime.

artrmz:
@cardoso @rafaelks Should be working now. My mistake by setting wrong default value of user's list cell.

Rocket.Chat/NewRoomViewController.swift
```diff
@@ -120,9 +135,13 @@ class NewRoomViewController: BaseViewController {
...
 
     @IBAction func buttonCreateDidPressed(_ sender: UIButton) {
-        guard let roomName = setValues["room name"] as? String else { return }
-        guard let publicRoom = setValues["public room"] as? Bool else { return }
-        guard let readOnlyRoom = setValues["read only room"] as? Bool else { return }
+        guard
+            let roomName = setValues["room name"] as? String,
+            let publicRoom = setValues["public room"] as? Bool,
+            let membersRoom = setValues["users list"] as? [String],
+            let readOnlyRoom = setValues["read only room"] as? Bool else {
```
cardoso: else in new line

```swift
guard
    let ...
    let ...
else {
    return
}
```

Rocket.Chat/Views/Form/Cells/MentionsTextFieldTableViewCell.swift
```diff
@@ -0,0 +1,134 @@
...
+            let item = filteredResults[itemPosition]
+            self.addUserToInviteList(name: item.title)
+            self.textFieldInput.text = nil
+            self.delegate?.updateDictValue(key: self.key ?? "", value: self.invitedUsers())
+        }
+    }
+
+    private func invitedUsers() -> [String] {
+        var invitedUsers = [String]()
+        for user in users {
```
cardoso: Let's just do instead:
```swift
    private func invitedUsers() -> [String] {
        return users.map { $0.key }
    }
```

Rocket.Chat/API/Requests/UsersListRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Artur Rymarz on 12.11.2017.
+//  Copyright © 2017 Rocket.Chat. All rights reserved.
+//
+
+import SwiftyJSON
+
+typealias UsersListResult = APIResult<UsersListRequest>
+
+class UsersListRequest: APIRequest {
```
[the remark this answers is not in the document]
artrmz: Will change that, although it will be not possible to invite users who are not in local database. For servers with lots of channels it will be prettty common issue I think?

artrmz:
@cardoso Pushed all your suggestions

cardoso:
Thanks @artrmz ! 🙌 

We'll merge this soon for 1.8

Rocket.Chat/NewRoomViewController.swift
```diff
@@ -193,6 +213,24 @@ extension NewRoomViewController: FormTableViewDelegate {
     func getPreviousValue(key: String) -> Any? {
         return setValues[key]
     }
+
+    func updateTable(key: String) {
+        tableView.beginUpdates()
+        tableView.endUpdates()
```
rafaelks: What does these two lines make? 🤔

rafaelks:
Just resolved the conflicts with `develop`. Let's merge it. Thank you very much @artrmz!

Rocket.Chat/NewRoomViewController.swift
artrmz: Forces tableView to recalculate cells height with animation

rafaelks:
Looking good! Let's merge it! 👍

PR #969, opened 2017-11-16
[NEW] Use UserNotifications API & open channel on notification tap

### LINKED ISSUE #790: [IMPROVEMENT] Push notifications - open proper channel + deprecated API
After tapping on push notification we should be redirected to channel for which the notification has been sent.

Also `UIApplicationDelegate` notification' methods are deprecated since IOS 10. We should use new framework called `UserNotifications`.

cardoso:
@RocketChat/ios

# Description

This PR aims to: 
- Upgrade our notification handling to the new APIs
- Open correct server and channel when tapping notification

Closes #790 

## Progress

- [x] Upgrade our notification handling to the new APIs
- [x] Show notifications when in foreground (happens when in different server)
- [x] Open correct server and channel on notification tap (suspended, background & foreground)
- [x] Tests


rafaelks:
This is looking super nice @cardoso! Thank you!

PR #985, opened 2017-11-22
[FIX] Avoid accessing invalidated objects in some places

### LINKED ISSUE #956: [CRASH] RLMObject_Private.hpp line 45 ___ZN12_GLOBAL__N_115makeBoxedGetterIN5realm10StringDataEEEP
# Information

**Introduced in:** 1.1.4 (69)
**OS:** iOS 10, iOS 11
**Device:** Any device
**Occurrences so far (from 90 days ago):** 80

# Stacktrace

```
Fatal Exception: RLMException
0  CoreFoundation                 0x1831efd38 __exceptionPreprocess
1  libobjc.A.dylib                0x182704528 objc_exception_throw
2  Realm                          0x103547e58 ___ZN12_GLOBAL__N_115makeBoxedGetterIN5realm10StringDataEEEP11objc_objectm_block_invoke (RLMObject_Private.hpp:45)
3  Rocket.Chat                    0x102f8c0d0 specialized ChatViewController.collectionView(UICollectionView, layout : UICollectionViewLayout, sizeForItemAt : IndexPath) -> CGSize (ChatViewController.swift:24)
4  Rocket.Chat                    0x102f8202c @objc ChatViewController.collectionView(UICollectionView, layout : UICollectionViewLayout, sizeForItemAt : IndexPath) -> CGSize (ChatViewController.swift)
5  UIKit                          0x18d0a1fa8 -[UICollectionViewFlowLayout _getSizingInfosWithExistingSizingDictionary:]
6  UIKit                          0x18d0a3538 -[UICollectionViewFlowLayout _fetchItemsInfoForRect:]
7  UIKit                          0x18c75d730 -[UICollectionViewFlowLayout prepareLayout]
8  UIKit                          0x18c65b000 -[UICollectionViewData _prepareToLoadData]
9  UIKit                          0x18d081edc -[UICollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:animator:]
10 UIKit                          0x18d0864d4 -[UICollectionView _endUpdatesWithInvalidationContext:tentativelyForReordering:animator:]
11 UIKit                          0x18d0867b8 -[UICollectionView _performBatchUpdates:completion:invalidationContext:tentativelyForReordering:animator:]
12 UIKit                          0x18d086614 -[UICollectionView _performBatchUpdates:completion:invalidationContext:tentativelyForReordering:]
13 UIKit                          0x18d086598 -[UICollectionView _performBatchUpdates:completion:invalidationContext:]
14 UIKit                          0x18c7bd6c0 -[UICollectionView performBatchUpdates:completion:]
15 Rocket.Chat                    0x102f88f6c specialized ChatViewController.(loadHistoryFromRemote(date : Date?) -> ()).(closure #1).(closure #1) (ChatViewController.swift:575)
16 Rocket.Chat                    0x102f8a638 partial apply for ChatViewController.(loadHistoryFromRemote(date : Date?) -> ()).(closure #1).(closure #1) (ChatViewController.swift)
17 Rocket.Chat                    0x102f7c224 thunk (ChatViewController.swift)
18 libdispatch.dylib              0x182b75088 _dispatch_call_block_and_release
19 libdispatch.dylib              0x182b75048 _dispatch_client_callout
20 libdispatch.dylib              0x182b81b74 _dispatch_main_queue_callback_4CF$VARIANT$mp
21 CoreFoundation                 0x183197f20 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
22 CoreFoundation                 0x183195afc __CFRunLoopRun
23 CoreFoundation                 0x1830b62d8 CFRunLoopRunSpecific
24 GraphicsServices               0x184f47f84 GSEventRunModal
25 UIKit                          0x18c663880 UIApplicationMain
26 Rocket.Chat                    0x102ee5b4c main (ServerManager.swift:14)
27 libdyld.dylib                  0x182bda56c start
```

rafaelks:
@RocketChat/ios

Closes #956


Rocket.Chat/Views/Avatar/AvatarView.swift
```diff
@@ -24,7 +24,13 @@ final class AvatarView: UIView {
...
     var user: User? {
         didSet {
-            updateAvatar()
+            if let user = user {
+                if !user.isInvalidated {
+                    updateAvatar()
+                }
+            } else {
+                updateAvatar()
+            }
```
cardoso: How about we revert this and change the userAvatarURL() function instead?
That way we don't crash if we call it from somewhere else.

```swift
private func userAvatarURL() -> URL? {
    guard let user = user, !user.isInvalidated,
          let username = user.username,
          let auth = AuthManager.isAuthenticated(),
          let baseURL = auth.baseURL(),
          let encodedUsername = username.addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed)
    else {
        return nil
    }

    return URL(string: "\(baseURL)/avatar/\(encodedUsername)")
}
```

rafaelks: @cardoso The problem with that is that we also use `user` object when we need to set the initials.

cardoso: @rafaelks I'd just check there too. I think it's more maintainable if we add the check as pre-condition to the functions instead of before calling them, that way we can never forget.

Also would be better if we made the functions in this class more testable like
```swift 
func avatarWithInitialsForString(_ string: String) -> (backgroundColor: UIColor, text: String)? {
    ... // don't need user here
}
```
But that's just a suggestion, we can refactor this class later :)

PR #1004, opened 2017-11-24
[NEW] Show admin panel for users that has access to it

### LINKED ISSUE #961: [NEW] Let user open admin panel (if user has access to it)

rafaelks:
@RocketChat/ios

Closes #961


Rocket.Chat/SettingsViewModel.swift
```diff
@@ -27,6 +27,14 @@ final class SettingsViewModel {
...
     }
 
+    internal var canViewAdminPanel: Bool {
+        guard let user = AuthManager.currentUser(), !user.isInvalidated else { return false }
+        return user.hasPermission(.viewPrivilegedSetting) ||
+            user.hasPermission(.viewStatistics) ||
+            user.hasPermission(.viewUserAdministration) ||
+            user.hasPermission(.viewRoomAdministration)
+    }
+
```
cardoso: We can test this if we make it a method or computed variable of User

user.canViewAdminPanel

Rocket.Chat/Controllers/WebViewEmbedded/WebViewControllerEmbedded.swift
```diff
@@ -0,0 +1,48 @@
...
+//  Rocket.Chat
+//
+//  Created by Rafael Kellermann Streit on 11/24/17.
+//  Copyright © 2017 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+import WebKit
+
+class WebViewControllerEmbedded: UIViewController {
```
cardoso: Can you add an ActivityIndicator like we do in OAuthViewController.swift? It stays white for a long time here.

rafaelks:
@cardoso Ready to review again! 👍

PR #1022, opened 2017-11-30
[FIX] Hide Back Button if a Multiserver Support is disabled

### LINKED ISSUE #1021: [BUG] Back Button in Login View Crashed the App if Multiserver Config is disabled
- Your Rocket.Chat app version: 1.7.3
- Your Rocket.Chat server version: 0.59.3
- Device (or Simulator) you're running with: iPhone

If you set disable the Multiserver Config by setting a fixed hostname in the plist.info the app successfully skips the screen where this url is entered, but on the login screen the back button is still visible. If you then click the back button the app crashes.

janrenz:
@RocketChat/ios
This PR removes the back button in case Multiserver Support is disabled.

Closes #1021

### DIFF
--- a/Rocket.Chat/Controllers/Auth/AuthViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/AuthViewController.swift
@@ -66,6 +66,10 @@ final class AuthViewController: BaseViewController {
         if let registrationForm = AuthSettingsManager.shared.settings?.registrationForm {
            buttonRegister.isHidden = registrationForm != .isPublic
         }
+        
+        if !AppManager.supportsMultiServer {
+            self.navigationItem.hidesBackButton = true
+        }
 
         self.updateAuthenticationMethods()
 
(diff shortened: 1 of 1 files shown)


jprberlin:
HI, I saw that code coverage has decreased slightly. However the full controller is not tested. The new feature would be best tested by an UI test (sure back button is visible by default), but they are none so far.. So what should I do?

rafaelks:
@jprberlin Don't worry, we're not testing the UI right now, so that's OK... we'll review your PR soon. Thank you!

rafaelks:
@jprberlin I cannot reproduce the crash that you're mentioning... what are the steps to reproduce? The problem in removing the "Back" button is that we're currently using the `ConnectServerViewController.swift` to connect into the server. So if user lose the connection with the WebSocket in the Auth screen, it will be required to go back to the first screen and connect again into the server. 🤔

rafaelks:
Closing this issue per discussion on #1021. Let's merge it when we make sure user won't be blocked on the screen if the app gets disconnected.

PR #1037, opened 2017-12-05
[CHORE] Replace deprecated pod Google/SignIn with GoogleSignIn

### LINKED ISSUE #1030: [CHORE] Replace deprecated Google pod
Replace deprecated pod `Google/SignIn` with the newer pod `GoogleSignIn`.

*Questions:*
- Do we use Firebase? If so should we also include `pod 'Firebase/Core'` since it is not delivered with `GoogleSignIn` as it was with `Google/SignIn`
- Do we use `GoogleAnalytics`? If so -> `pod 'GoogleAnalytics'`

artrmz:
@RocketChat/ios

Closes #1030


Rocket.Chat/Controllers/Auth/AuthViewControllerGoogleExtensions.swift
```diff
@@ -6,12 +6,13 @@
...
 
+import GoogleSignIn
+
 extension AuthViewController {
     func authenticateWithGoogle() {
         startLoading()
 
-        var configureError: NSError?
-        GGLContext.sharedInstance().configureWithError(&configureError)
+        GIDSignIn.sharedInstance().clientID = "662030055877-9qc3lqsbgif4k1ktl89dqeaec3m163i4.apps.googleusercontent.com"
```
artrmz: I don't like this change but seems like with newer library we have to set it manually instead of calling any `configure` method like it was before.

artrmz:
@rafaelks Seems like tests failed because of outdated pod master repo on CI. Any way to `pod repo update` on CI?

rafaelks:
@artrmz Thank you! We're finishing a release process of 1.7.4 and then we're going to review this. 👍

rafaelks:
@artrmz This command is suppose to do that: https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Scripts/update_dependencies.sh#L9.

Maybe we can change it to `pod repo update`?

Rocket.Chat/Controllers/Auth/AuthViewControllerGoogleExtensions.swift
rafaelks: 🙈

rafaelks: What about the `GoogleService-Info.plist` file?

artrmz:
@rafaelks If we want to be independent then I think so. Seems like this script updated pods repo from some other server instead of Git to speed it up but it's not always (rare case?) up to date

rafaelks:
@artrmz Looks like it. Do you mind updating that line in your PR so we give it a try?

artrmz:
@rafaelks Tests have passed this time

rafaelks:
@artrmz Great! I made a few comments on the code. Can you take a look?

Rocket.Chat/Controllers/Auth/AuthViewControllerGoogleExtensions.swift
artrmz: According to documentation https://developers.google.com/identity/sign-in/ios/start-integrating it should be included.

rafaelks: But if we're using the file, we can keep using the configureWithError(), like here: https://developers.google.com/identity/sign-in/ios/sign-in?ver=swift.

artrmz: There is no GGLContext class anymore

artrmz: I have removed it and did some tests, seems like it's not needed. Logging with Google account works without this file.

rafaelks:
@artrmz Thank you! Let's merge it and keep testing it with different servers until we release it.

PR #1048, opened 2017-12-07
[NEW] Edit message

luismachado:
@RocketChat/ios

Adds the ability to edit a previously sent message


luismachado:
I'm a bit confused by the failure on the ipad test. From what I've managed to see it's a timed test which also seemed to fail (> 1 second) on my previous commits despite it now being shown.

rafaelks:
@luismachado I'll run it again, thanks!

luismachado:
It looks like the 1.8.0 release will be pushing REST API usage for the send message operation.
To update the messages is the endpoint "/api/v1/chat.updateMessage" correct?

It would probably be a good idea to have the URLs specified in a different file no?

rafaelks:
@luismachado You're right, we're pushing to use REST APIs. I think the API you're looking for is this one: https://rocket.chat/docs/developer-guides/rest-api/chat/update/.

Yes, at some point we'll have a file with all the URLs as constants and use them in our code. 👍

rafaelks:
Feature is working as expected, but can you use the **Edit Mode** we have on the **SlackTextViewController**? You can see it here: https://github.com/slackhq/SlackTextViewController/blob/37f91f23414dda33db5ce43a20958f19ddbd7b1d/Source/SLKTextViewController.m#L777.

luismachado:
@rafaelks I'll look into that, however there's an issue that should be addressed first.

I just pulled the latest commits from 'development' (into a separate branch that I created) and I noticed that, after that, when I edit a message it's still sent correctly to the server (I can see it on the webapp) but it isn't changing in the app itself. Any idea of what might be happening?

Two more things:
- When the app is started the chat it opens in isn't automatically refreshing. You have to select another chat and then get back to the original one to automatically show new messages, show the "user is typing.." message. This isn't a recent issue. Were you aware of it?

- With the latest changes, if a message is edited and then the other user sends a message the app crashes (in `updateMessagesQueryNotificationBlock` when it calls `self.appendMessages`). This may be correlated with the issue of not updating the edited message no?

Let me know if you need any further info!

rafaelks:
@luismachado Good finding! Just to make sure, what server version are you using? Because I know we had some issues related to it when using 0.60.rc.0. Are you using `open.rocket.chat`? If yes, let's try to reproduce this issues again.

luismachado:
Hello @rafaelks . Thanks for getting back at this.
I don't have access to the admin part of the server (as I'm only working on the iOS app) so I won't be able check the version right now; unless I can check it anywhere else? Otherwise, I'll check it as soon as possible.

So, I'm not using open.rocket.chat but next week I can try pointing my version there and try to reproduce the issues. Are you talking about a specific one by the way? (edited msg not refreshing / new messages not appearing / crash when receiving msg)

rafaelks:
@luismachado You can check the server version by accessing this endpoint:

```shell
curl "https://open.rocket.chat/api/info" | python -mjson.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   218  100   218    0     0    304      0 --:--:-- --:--:-- --:--:--   304
{
    "success": true,
    "version": "0.60.0-rc.2"
}
```

I'm talking about the edit message not refreshing and new messages not appearing. The crash is possible that's not related to the server version. Did you manage to get any stacktrace of it?

luismachado:
Hey,

The version is:
`  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    44    0    44    0     0    247      0 --:--:-- --:--:-- --:--:--   248
{
    "success": true,
    "version": "0.58.4"
}`

Regarding the stack trace, I'll reproduce the issue next Monday and I'll try to get the stacktrace ;)

rafaelks:
@luismachado Hum... that's weird, this issue should not happen with this server version 🤔.

Great, I'll wait for your logs.

luismachado:
Okay so after #1078 the issue where the messages weren't showing on the first chat is now solved!

However the other two persist:
- Edit text: edited text doesn't change on the chat (on the web client it shows the modification)
- After the text is edited, if the other member in the chat sends a message the app crashes. For some reason no stack trace is displayed (never had this happen to me before, maybe something related with Meteor?). I took a screenshot however.

Considerations:
- This worked on the 8th, so it probably is some code change after that that's causing this.
- The issue here seems to be that, when the message is edited via iOS that message gets, somehow, broken:
      - If I edited it again on the web app the change WON'T reflect on the app
      - However If I edit any other message on the web client that message WILL BE changed on the app
      - If I change chats and then change back to the original then everything will be okay and the edited message will now display the changes (and if I edit it on the web client it will show those changes)
      - The crash issue only occurs AFTER editing a message and if I remain on that chat. If I change and then change back it will receive the new messages correctly.
      - Could this be some issue with the 'subscription'? I don't know that part of the code at all.
      - I noticed that the send msg is now using REST API (with websocket fallback). Could that be it? The edit procedure still uses the [websocket](https://github.com/RocketChat/Rocket.Chat.iOS/pull/1048/files#diff-914164611cb69f4158d88d4fb8927685R410). However on my tests I used both our own server (which is a older version so it will fallback to the websocket) and the open.rocket.chat so it should be that :/

luismachado:
Okay! @rafaelks I managed to find out the [issue](https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.Chat/Controllers/Chat/ChatDataController.swift#L209).

Removing that condition fixes both the edited message not updating and the app will also not crash afterwards. Looking the git blame it looks like it was you @rafaelks that added this so I wanted to check if this is really necessary or/if there's a way to fixing the bug without removing this line!

luismachado:
Okay @rafaelks I pull the most recent changes into my branch, comment [this](https://github.com/RocketChat/Rocket.Chat.iOS/pull/1048/files#diff-bb46f58694cfd8491fdea9c9f7e5c1c5R209) to fix the issues explained above (please check if that is okay or what we can do differently) and now I'm also using the edit mode from slack view :)

rafaelks:
@luismachado Thank you, I'll test your changes. Do you mind fixing Hound comments also?

rafaelks:
@luismachado Edit feature is looking great, but I cannot send messages anymore? Haha! On my tests, I cannot send any message, just edit.

luismachado:
@rafaelks Okay, it is now working! Was missing a call to the sendTextMessage method -.-' (which, in the send audio message pull request wasn't needed, hence me forgetting it).

Rocket.Chat/API/Clients/MessagesClient.swift
```diff
@@ -47,7 +47,7 @@ struct MessagesClient: APIClient {
             switch error {
             case .version:
                 // TODO: Remove SendMessage Fallback + old methods after Rocket.Chat 1.0
-                SubscriptionManager.sendTextMessage(message, completion: { response in
+                SubscriptionManager.sendEditTextMessage(message, completion: { response in
```
rafaelks: Why did you rename this method? I guess we can have both.

Rocket.Chat/Controllers/Chat/ChatControllerReplyHandler.swift
```diff
@@ -15,7 +15,7 @@ extension ChatViewController {
         replyView.backgroundColor = textInputbar.addonContentView.backgroundColor
         replyView.frame = textInputbar.addonContentView.bounds
         replyView.onClose = stopReplying
-
+        replyView.alpha = 0
```
rafaelks: Why did you add this line?

Rocket.Chat/Controllers/Chat/ChatControllerReplyHandler.swift
```diff
@@ -24,24 +24,29 @@ extension ChatViewController {
...
+        
         textView.becomeFirstResponder()
 
         replyString = (onlyQuote ? message.quoteString : message.replyString) ?? ""
 
         scrollToBottom()
     }
 
+    func clearReplying() {
+        self.replyView.alpha = 0
```
rafaelks: There's no need to `self.` here I guess.

Rocket.Chat/Controllers/Chat/ChatDataController.swift
```diff
@@ -206,9 +206,9 @@ final class ChatDataController {
     func update(_ message: Message) -> Int {
         for (idx, obj) in data.enumerated()
             where obj.message?.identifier == message.identifier {
-                if obj.message?.updatedAt?.timeIntervalSince1970 == message.updatedAt?.timeIntervalSince1970 {
-                    return -1
-                }
+                //if obj.message?.updatedAt?.timeIntervalSince1970 == message.updatedAt?.timeIntervalSince1970 {
+                //   return -1
+                //}
```
rafaelks: Wondering why did you comment this line... this will cause tons of updates in cells that doesn't need to be updated.

Rocket.Chat/Managers/Model/SubscriptionManager.swift
```diff
@@ -353,10 +353,13 @@ extension SubscriptionManager {
...
 
-    static func sendTextMessage(_ message: Message, completion: @escaping MessageCompletion) {
+    static func sendEditTextMessage(_ message: Message, isEdit: Bool = false, completion: @escaping MessageCompletion) {
+
+        let methodName: String = isEdit ? "updateMessage" : "sendMessage"
+
         let request = [
             "msg": "method",
-            "method": "sendMessage",
+            "method": methodName,
```
rafaelks: Let's have two separated methods here instead of changing the method name based on `isEdit`. OK?

Rocket.Chat/cs.lproj/Localizable.strings
```diff
@@ -106,13 +106,15 @@
 "chat.dm.start_conversation" = "Toto je začátek vaší konverzace s %@";
 "chat.channel.start_conversation" = "Toto je začátek konverzace %@";
 "chat.read_only" = "Tato místnost je pouze pro čtení";
-"chat.muted" = "You have been muted"; //TODO
+"chat.muted" = "You have been muted";
+"chat.edit_mode" = "Upravit zprávu. Blízko zrušit.";
```
rafaelks: Could you add the strings to the same places in all the languages file? This makes easier to maintain in the future.

Rocket.Chat/Controllers/Chat/ChatControllerReplyHandler.swift
luismachado: Hum this was needed before I started using SlackTextViewController's EditMode (as I had an editView and a replyView and had to manage their visibilities). It isn't needed now, good catch!

Rocket.Chat/Controllers/Chat/ChatDataController.swift
luismachado: This is what was causing the issue where edited things would not be refreshed on the app and, afterwards, receiving a message would cause a crash. Please check my post from 22 days ago (near commit 4ae3eb6)

rafaelks: Yeah, I can see that... but we can't remove this verification. You need to update the `updatedAt` value of your message in order to update the cell. Makes sense?

luismachado: It does, but I already do that in [here](https://github.com/RocketChat/Rocket.Chat.iOS/pull/1048/files#diff-914164611cb69f4158d88d4fb8927685R409).

Maybe this should be done in other place?

rafaelks: Well... have you tried uncommenting this code and trying to edit a message?

rafaelks:
REST: https://rocket.chat/docs/developer-guides/rest-api/chat/update/

cardoso:
Thanks for the PR @luismachado 

I made a few code improvements, added tests and now it's using the REST API.

It's ready to review again @rafaelks

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -375,6 +396,10 @@ final class ChatViewController: SLKTextViewController {
         }
     }
 
+    override func textViewDidEndEditing(_ textView: UITextView) {
+
+    }
```
rafaelks: Is this code required?

rafaelks:
Awesome @cardoso & @luismachado, thank you guys! 👊

PR #1067, opened 2017-12-12
[CHORE] Improve the performance of the app, reducing calls on main-thread and the number of reutilization of the cells

### LINKED ISSUE #1064: [BUG] Something is blocking the UI of the app on reconnecting to the server

rafaelks:
@RocketChat/ios

Closes #1064 

## Description

This PR fixes a few performance and bugs in the app:

1. ImageView flickering on the messages list;
2. Cell was being reused hundreds of times on any Message update call from Realm;
3. Moved the Markdown processing to the main thread again, to avoid weird behaviors;

### DIFF
--- a/Rocket.Chat/Managers/Model/AuthSettingsManager.swift
+++ b/Rocket.Chat/Managers/Model/AuthSettingsManager.swift
@@ -38,7 +38,7 @@ final class AuthSettingsManager {
                 return
             }
 
-            Realm.executeOnMainThread({ realm in
+            Realm.execute({ realm in
                 let settings = AuthManager.isAuthenticated()?.settings ?? AuthSettings()
                 settings.map(response.result["result"], realm: realm)
                 realm.add(settings, update: true)
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageTextViewModel.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageTextViewModel.swift
@@ -43,9 +43,9 @@ final class ChatMessageTextViewModel {
         self.attachment = attachment
     }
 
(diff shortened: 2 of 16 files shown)


rafaelks:
@cardoso This is ready to review!

Rocket.Chat/Managers/Model/AuthSettingsManager.swift
```diff
@@ -38,7 +38,7 @@ final class AuthSettingsManager {
                 return
             }
 
-            Realm.executeOnMainThread({ realm in
+            Realm.execute({ realm in
                 let settings = AuthManager.isAuthenticated()?.settings ?? AuthSettings()
                 settings.map(response.result["result"], realm: realm)
                 realm.add(settings, update: true)
```
cardoso: Just crashed here 😦 "not in a write transaction" when changing servers.

cardoso: Looks like this causes very bad problems @rafaelks, can we revert only this part? Can't even connect to servers with this.

Rocket.Chat/Views/Cells/Chat/ChatMessageTextViewModel.swift
```diff
@@ -43,9 +43,9 @@ final class ChatMessageTextViewModel {
         self.attachment = attachment
     }
 
-    func toggleCollpase() {
-        Realm.executeOnMainThread({ _ in
+    func toggleCollpase(_ completion: VoidCompletion? = nil) {
+        Realm.execute({ _ in
             self.attachment.collapsed = !self.attachment.collapsed
```
cardoso: Crashing here too

PR #1078, opened 2017-12-16
[FIX] Messages not updating after opening app or changing servers

cardoso:
@RocketChat/ios

We were subscribing before we had a connection. (Another bug squashed from our optimization :))


luismachado:
This might fix the issues that I encounter while working on my PR #1048 @rafaelks .
I'll keep that on hold until this PR is merged and then I'll check if the issues there are gone ;)

cardoso:
Thanks for taking interest @luismachado. This was also troubling me while working on reactions.
 I'll resume the work and see if we can solve this today.

PR #1061, opened 2017-12-11
[NEW] Support sending audio records

luismachado:
@RocketChat/ios

I forked #762 (as no update was given since a month ago)  so credit also to @falcaoaugustos (maybe it's best to update his PR with this one?)

Changes since the fork:
- Visual feedback is different now (whole screen is blurred while recording and a timer is also present)
- Feedback is now quicker as I replaced the button action for a tapgesturerecognizer (`delayTouchesBegan` was also removed)
- Code is also up to date with the most recent commits.

PR #1089, opened 2017-12-26
Revert "[NEW] Support sending audio records"

rafaelks:
Reverts RocketChat/Rocket.Chat.iOS#1061

I'll revert this PR for a while until we improve the icons and the responsiveness while recording an audio. Functionality looks good, but sometimes the app freezes and the icons aren't following the same pattern of the others.


rafaelks:
@luismachado I'm sorry about that, but we'll have to release 1.8.0 and this PR needs a little improvements. I'll try to update some stuff to release this right after 1.8.0.

PR #1090, opened 2017-12-26
[FIX] Check for title_link_download real value instead of always true

### LINKED ISSUE #1087: [BUG] Messages with titleLink are getting displayed wrongly
The app is understanding titleLink as a file to be downloaded. This isn't always the case.

rafaelks:
@RocketChat/ios

Closes #1087


cardoso:
Looks like in test-ipad-ios10.3 an async test failed? :thinking: need to re-run that. I'll take a look on adjusting some values later.

PR #1101, opened 2018-01-02
[FIX] Issue on connecting to old servers or trying to close connect screen

### LINKED ISSUE #1070: [BUG] Add Server screen after logout, even having other servers
We have this weird bug now:

When we log out, we're not switching to the first server's realm which causes the App to think there's no authenticated server. And then we get stuck in this screen (close button doesn't work)

rafaelks:
@RocketChat/ios

Closes #1070


cardoso:
Bug still happening here @rafaelks. Can you check again?

PR #1115, opened 2018-01-05
[FIX] Remove duplicated slashes from any URL that opens embedded in a WebView

### LINKED ISSUE #1111: [BUG] Admin page not loading
I just tried to open the admin page on our server with 0.60.2 in Build 132, but it just shows a spinning wheel and then a blank page.

### LINKED ISSUE #1118: [BUG] Sidebar doesn't update all values on logout
When logging out from one server and switching automatically to another server following params weren't updated:
 - Server Name (retaining old servers title)
 - Profile picture (set to default svg, neither updating to the current servers image nor retaining the old servers image)
 - Online status (offline)
So more or less the whole top elements except the username are not updated correctly in the sidebar.

rafaelks:
@RocketChat/ios

Closes #1111 
Closes #1118

### DIFF
--- a/Rocket.Chat/Extensions/URLExtension.swift
+++ b/Rocket.Chat/Extensions/URLExtension.swift
@@ -81,4 +81,14 @@ extension URL {
         return newURL
     }
 
+    mutating func removingDuplicatedSlashes() -> URL {
+        let urlString = self.absoluteString.replacingOccurrences(of: "//", with: "/")
+
+        if let newURL = URL(string: urlString) {
+            self = newURL
+        }
+
+        return self
+    }
+
 }
--- a/Rocket.Chat/Extensions/URLExtension.swift
+++ b/Rocket.Chat/Extensions/URLExtension.swift
@@ -81,17 +81,12 @@ extension URL {
         return newURL
     }
 
-    mutating func removingDuplicatedSlashes() -> URL {
+    func removingDuplicatedSlashes() -> URL? {
         let newURLPath = self.path.replacingOccurrences(of: "//", with: "/")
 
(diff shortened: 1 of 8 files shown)


Rocket.Chat/Extensions/URLExtension.swift
```diff
@@ -81,4 +81,14 @@ extension URL {
         return newURL
     }
 
+    mutating func removingDuplicatedSlashes() -> URL {
```
cardoso: Can you write tests for this @rafaelks? Does it work with https:// ?

rafaelks: Done! :-)

cardoso: Sorry for not catching this before, but does this method need to be mutating? If it must be mutating then we should rename it to "removeDuplicatedSlashes".

soundstorm:
@rafaelks but this will break some URLs, right? 😢
e.g. https://www.aliexpress.com/item//32798469203.html (ok) and https://www.aliexpress.com/item/32798469203.html (404)

The main issue was the server url containing a trailing slash, so can't we just apply a stripper of the trailing slash if there is any?

cardoso:
Then we should just advise the admin to remove the trailing slash? @rafaelks @soundstorm

rafaelks:
@soundstorm @cardoso Yeah... I guess we can leave the app without this change for now. Thanks guys!

soundstorm:
@cardoso I logged out, entered the url again without trailing slash but it fetched the //admin again (?)

rafaelks:
Let's merge this one, for the better.

rafaelks:
@cardoso This is ready to review again.

PR #1131, opened 2018-01-08
[NEW] Integrate FLEX in debug mode

### LINKED ISSUE #821: [NEW] Add debugging to debug/ad-hoc builds
Need to find a few suggestions.

cardoso:
@RocketChat/ios

Press F on simulator to show the FLEX view

More details at https://github.com/Flipboard/FLEX

Closes #821


Rocket.Chat/Controllers/Chat/ChannelInfoViewController.swift
```diff
@@ -22,6 +23,14 @@ class ChannelInfoViewController: BaseViewController {
         didSet {
             guard let subscription = self.subscription else { return }
 
+            #if DEBUG
+                let flex = [ChannelInfoDetailCellData(title: localized("Open FLEX"), detail: "", action: {
```
rafaelks: 1. String is not localized (and doesn't need to be), just use a String and add **DEBUG: Open FLEX**;
2. Let's also add this on **BETA**;
3. Let's move it to Settings;

PR #1163, opened 2018-01-13
[I18N] Fix fr translation

### LINKED ISSUE #1162: [I18N] [BUG] No more localization
- Your Rocket.Chat+ app version: 1.8.0 (137) and previously 1.7 (126)
- Your Rocket.Chat server version: 0.60.2
- Device (or Simulator) you're running with: Iphone IOS 11.1.1

My Rocket.chat server is in french. My iphone is in french, my rocket.chat+ and rocket.chat apps were in french.
I made an upgrade from 1.7 (126) to 1.8.0 (137)... and now rocket.chat+ app is in english (and i can't modify as there is no menu for that) , while rocket.chat app remains in french.

Would it be possible to have the localization ?

soundstorm:
@RocketChat/ios

Closes #1162


soundstorm:
@rafaelks maybe we need a 1.8.1 (don't know how many french users we have)

cardoso:
Thanks for the :fr: translations @soundstorm! 

I pushed some tweaks to keep the lines consistent between localization files.

PR #1172, opened 2018-01-17
[I18N] Translate quick reply

soundstorm:
@RocketChat/ios

Although we have no other option on mobile and thus just show the keyboard to reply instantly, on Watch we see the "Reply" action button


rafaelks:
Thank you @soundstorm! 👍

PR #1179, opened 2018-01-19
[FIX] Sometimes subscriptions going out of sync

### LINKED ISSUE #1086: [BUG] Check if it's possible to force a new sync every time user opens the app
- Your Rocket.Chat app version: 1.7.4 (126)
- Your Rocket.Chat server version: https://chat.zcashcommunity.com/
- Device (or Simulator) you're running with: iPhone 6s

Steps:
1. Log in as A from device A.
2. Connect to a server A on device A.
3. Add some channels on device A.
4. Log in as A from device B.
5. Connect to server A on device B.

What happened:
Not all channels are synced.

Expected result:
All channels are synced.

cardoso:
@RocketChat/ios

Closes #1086

### DIFF
--- a/Rocket.Chat/Managers/Model/SubscriptionManager.swift
+++ b/Rocket.Chat/Managers/Model/SubscriptionManager.swift
@@ -46,6 +46,11 @@ struct SubscriptionManager {
             SocketManager.send(requestRooms) { response in
                 guard !response.isError() else { return Log.debug(response.result.string) }
 
+                guard let auth = AuthManager.isAuthenticated() else { return }
+                Realm.executeOnMainThread { _ in
+                    auth.lastSubscriptionFetch = Date.serverDate
+                }
+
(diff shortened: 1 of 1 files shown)


rafaelks:
Good finding! I think that makes sense! 👍

Rocket.Chat/Managers/Model/SubscriptionManager.swift
```diff
@@ -46,6 +46,11 @@ struct SubscriptionManager {
             SocketManager.send(requestRooms) { response in
                 guard !response.isError() else { return Log.debug(response.result.string) }
 
+                guard let auth = AuthManager.isAuthenticated() else { return }
```
soundstorm: @cardoso @rafaelks 
When adding a new server this results in
```
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```
So App crashes everytime adding a new server! 😢

PR #1091, opened 2017-12-28
[FIX] Remove size information when uploading a file (to prevent mistakes)

rafaelks:
@RocketChat/ios

Closes #1075

PR #1197, opened 2018-01-24
Revert "[FIX] Remove size information when uploading a file (to prevent mistakes)"

rafaelks:
Reverts RocketChat/Rocket.Chat.iOS#1091

### DIFF
--- a/Rocket.Chat/Helpers/UploadHelper.swift
+++ b/Rocket.Chat/Helpers/UploadHelper.swift
@@ -22,6 +22,7 @@ struct UploadHelper {
     static func file(for data: Data, name: String, mimeType: String) -> FileUpload {
         return FileUpload(
             name: name,
+            size: sizeFor(data),
             type: mimeType,
             data: data
         )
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Helpers/UploadHelper.swift
```diff
@@ -45,6 +46,17 @@ struct UploadHelper {
...
 
+    /**
+        This method returns the size of some Data object.
+
+        - parameters:
+          - data: The data object to be inspected.
+        - returns: The size of the data
+     */
+    static func sizeFor(_ data: Data) -> Int {
+        return (data as NSData).length
```
soundstorm: Seems to be working again - still on FileSystem. 🤔
Also tested directly on my phone: MongoDB size matches the file size. But as I updated the server to 0.60 on 28/12 that may have fixed the problem... So maybe it was just the server messing up.

rafaelks: @soundstorm The app is not using WebSockets to send the file when server is 0.60, that's why it works just fine everything, hehe.

PR #1203, opened 2018-01-25
[UPDATE] Update Starscream to 3.0.4

rafaelks:
@RocketChat/ios


Rocket.Chat/Managers/Socket/SocketManager.swift
```diff
@@ -182,7 +181,7 @@ extension SocketManager {
 
 // MARK: WebSocketDelegate
```
soundstorm: suggestion: change this to Advanced too?

PR #1191, opened 2018-01-23
[NEW] Image description

RahulDhiman93:
@RocketChat/ios

Closes #1166 
@cardoso

PR #1216, opened 2018-01-30
[NEW] Attachment description

### LINKED ISSUE #1170: [NEW] Show image description
The title of the image shows up, but we could show the description below the ChatMessageImageView in chat like a cite (variable height)

soundstorm:
Closes #1170

In addition closes #1191, as the solution was unfortunately not working.

Adds descriptions to images, videos and audio.

Quotemarks and description

@RocketChat/ios


Rocket.Chat/Models/Attachment.swift
```diff
@@ -29,6 +29,7 @@ class Attachment: BaseModel {
 
     @objc dynamic var collapsed: Bool = false
     @objc dynamic var text: String?
+    @objc dynamic var desc: String?
```
rafaelks: Could you change it to `descriptionText` please? Just to follow the same pattern we have in other models.

Rocket.Chat/Views/Cells/Chat/ChatMessageAttachmentView.swift
```diff
@@ -0,0 +1,24 @@
...
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+class ChatMessageAttachmentView: UIView {
+    class var defaultHeight: CGFloat {
+        return 0
+    }
+
+    static func heightFor(withText description: String?) -> CGFloat {
+        guard let text = description, text != "" else {
```
rafaelks: You can use `!text.isEmpty` also, instead of checking for `empty` string. It will prevent from allocating another String in memory.

PR #1217, opened 2018-01-30
[FIX] App hangs once in a while

### LINKED ISSUE #1177: [BUG] App hangs from time to time
I'm facing hanging of the UI sometimes - in most cases I want to scroll but can't. Rest of the UI is unresponsive too, but Instabug works (can move the Bubble around).
I'll try to reproduce the error with flex opened.

cardoso:
@RocketChat/ios

Greatly reduced usage of imageWithTint. Especially in the places it was causing hangs.

Closes #1177


rafaelks:
Sounds good @cardoso, let's try it! 👍

cardoso:
Seems like this improved it, but we still have some hangs... These are being harder to debug

PR #1233, opened 2018-02-07
[i18n] Localize missing polish strings

artrmz:
@RocketChat/ios

Localize polish strings


cardoso:
Thanks @artrmz! If you have the time, we also need translations for this file:

https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/pl.lproj/EmojiPicker.strings

artrmz:
@cardoso #1235

PR #1240, opened 2018-02-08
[HOTFIX] Decrease maximum size of LaunchScreen image

cardoso:
@RocketChat/ios 

LaunchScreen image was too big on iPad


rafaelks:
@cardoso I just made another commit, changing to 150px max, removing an unused constraint and adding a border constraint to phones as well, take a look please. 👍

PR #1213, opened 2018-01-28
[IMPROVEMENT] Update docs

soundstorm:
Updated docs, removed "pre-alpha", added how to contribute and guidelines, added screenshots

Closing #1044, Closing #1134, Closing #1135

PR #1241, opened 2018-02-08
Testing the skip of CI

soundstorm:
Testing #1213


soundstorm:
Okay, seems to be failing. Still linting with wip/branch

PR #1252, opened 2018-02-10
[TEST] CI Ignore

soundstorm:
I think wip/ and docs/ do not work as it shows up as user:wip/...


soundstorm:
But -wip does not work too. I'll test further on private repo :helpdesk_person:

PR #1273, opened 2018-02-16
[IMPROVEMENT] Prompt "App Store rating" after reacting to a message

### LINKED ISSUE #1264: [NEW] Prompt "App Store rating" after reacting to a message
We already have a code that handles that, just need to be called after reacting to a message.

artrmz:
@RocketChat/ios

Closes #1264

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
@@ -97,6 +97,7 @@ extension ChatViewController {
 
         controller.emojiPicked = { emoji in
             MessageManager.react(message, emoji: emoji, completion: { _ in })
+            UserReviewManager.shared.requestReview()
         }
 
         controller.customEmojis = CustomEmoji.emojis()
--- a/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
+++ b/Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
@@ -364,6 +364,8 @@ extension ChatMessageCell: UIGestureRecognizerDelegate {
 extension ChatMessageCell {
 
     override func awakeFromNib() {
+        super.awakeFromNib()
+        
         isAccessibilityElement = true
     }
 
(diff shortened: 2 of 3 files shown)


Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -140,6 +157,7 @@ final class ChatMessageCell: UICollectionViewCell {
         avatarView.prepareForReuse()
 
         for view in mediaViews.arrangedSubviews {
+            mediaViews.removeArrangedSubview(view)
```
artrmz: Not sure about this one, but I believe we should call it as well while preparing cell for reuse?

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -346,6 +364,8 @@ extension ChatMessageCell: UIGestureRecognizerDelegate {
 extension ChatMessageCell {
 
     override func awakeFromNib() {
+        super.awakeFromNib()
```
artrmz: According to docs we should pretty much always call it since many UIKit classes provides some additional actions there.

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
```diff
@@ -346,6 +364,8 @@ extension ChatMessageCell: UIGestureRecognizerDelegate {
...
+
```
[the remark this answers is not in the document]
artrmz: Just removed it in 4d72015. Seems like 2 quick commits in a row makes hound bark too much

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
rafaelks: Great finding!

Rocket.Chat/Views/Cells/Chat/ChatMessageCell.swift
rafaelks: Yes, you're right! Makes sense 👍

Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
```diff
@@ -97,6 +97,7 @@ extension ChatViewController {
 
         controller.emojiPicked = { emoji in
             MessageManager.react(message, emoji: emoji, completion: { _ in })
+            UserReviewManager.shared.requestReview()
```
rafaelks: 👍

rafaelks:
@artrmz Looks like the changes on the level of privacy of the vars broke the tests... I believe that instead of `private` you can use `internal`.

artrmz:
@rafaelks Oops, stupid me 😞 Will fix that in a moment

PR #1274, opened 2018-02-16
[i18n] Remove unused localize string

artrmz:
@RocketChat/ios

`chat.edit_mode` is not used anywhere


artrmz:
@rafaelks Was looking through already merged PR and noticed incorrect polish translation and was about to change the translation but noticed it wasn't used at all anyway ;)

rafaelks:
@artrmz Gotcha! Good finding, thank you! 👍

PR #1294, opened 2018-02-19
[FIX] Not moving from register screen after registration

### LINKED ISSUE #1284: [BUG] Not moving from register screen after successful registration

cardoso:
@RocketChat/ios

Some race condition problem

Closes #1284

### DIFF
--- a/Rocket.Chat/Controllers/Auth/SignupViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/SignupViewController.swift
@@ -130,12 +130,14 @@ final class SignupViewController: BaseViewController {
                 }
 
                 AuthManager.auth(email, password: password, completion: { _ in
-                    guard let user = AuthManager.currentUser() else { return }
-                    if user.username != nil {
-                        self?.dismiss(animated: true, completion: nil)
-                        AppManager.reloadApp()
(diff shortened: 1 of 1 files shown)


rafaelks:
@cardoso You mean, so the user will have to type email/username & password again?

cardoso:
I changed the approach @rafaelks

Rocket.Chat/Controllers/Auth/SignupViewController.swift
```diff
@@ -130,12 +130,14 @@ final class SignupViewController: BaseViewController {
                 }
 
                 AuthManager.auth(email, password: password, completion: { _ in
-                    guard let user = AuthManager.currentUser() else { return }
-                    if user.username != nil {
-                        self?.dismiss(animated: true, completion: nil)
-                        AppManager.reloadApp()
-                    } else {
-                        self?.performSegue(withIdentifier: "RequestUsername", sender: nil)
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
rafaelks: Why do we need this call? The Realm call inside of the method is suppose to be happening on the main thread. 🤔

PR #1295, opened 2018-02-19
[WIP] List of blocked users + unblock users

Hrayr2112:
@RocketChat/ios

I've added the method "unblockMessagesFrom" in MessageManager. 
Now we can tap on member cell and see Action Sheet with actions "Open Profile", "Unblock" and "Cancel". 

- 'Open Profile' opens the info page (Need add localization for "Open Profile")
- 'Unblock action' calls the method unblockMessagesFrom from MessageManager (Need add localization for "Unblock")

I think we shouldn't use long press gesture for this action, because there is a case:
User can forget who was blocked and don't use long press for more information about member. Therefore Action Sheet was added to didSelectRowAtIndexPath method. In this case, the user always knows about member's "unblock state". 

Please review my code. I know, there are some changes which should be added
I can use translator for localization, if you don't mind.


cardoso:
@Hrayr2112 some of your commits have different authors, "Macbook Home", "Macbook air"

You need to make sure the e-mails in all commits are the same as your github account.

There are many ways to change the emails of previous commits, this is one I found:
https://stackoverflow.com/questions/2919878/git-rewrite-previous-commit-usernames-and-emails

Hrayr2112:
I've just done the all actions which were in that instruction. Now we can see a lot of issues. Can you help solve this problem?

PR #1296, opened 2018-02-20
[CI] Speed up tests

rodrigok:
@RocketChat/ios

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,13 +7,13 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Update Dependencies
-          command: ./Scripts/update_dependencies.sh
+          name: Install Swiftlint
+          command: ./Scripts/install_swiftlint.sh
       - run:
           name: Lint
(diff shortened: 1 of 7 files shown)


circle.yml
```diff
@@ -7,13 +7,13 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Update Dependencies
-          command: ./Scripts/update_dependencies.sh
+          name: Install Swiftlint
+          command: ./Scripts/install_swiftlint.sh
```
artrmz: If we install dependencies before running SwiftLint we could just use the one downloaded via CocoaPods (and remove the install_swiftlint.sh)... if that's possible in circle.yml?

rafaelks: I think that's a good idea @artrmz. What do you think @rodrigok?

rodrigok: @rafaelks @artrmz it's a good idea since we use cache to save that and do not repeat the dependencies install on next steps

PR #1306, opened 2018-02-21
[IMPROVEMENT] Make spotlight case insensitive

### LINKED ISSUE #1265: [IMPROVEMENT] Search (channels, users) is case sensitive
- Your Rocket.Chat app version: 1.8.0
- Your Rocket.Chat server version: 0.61.0
- Device (or Simulator) you're running with: iPhone 5S

**Channel search is case sensitive**
Imagine you like to search for your channel `Team-chat`. If you insert `team-chat` in your channel search, Rocket.Chat+ will not find any match. Instead you have to search for the exact term (case sensitivity). Same applies if you search for users in the channel search. This is not user friendly. 

**Create new channel: User search is case sensitive** 
When you create a new channel and like to invite users, you have to insert the users' names. The search here is case sensitive as well. If you do not know the correct spelling, you simply will not find the users.

(edit: updated this issue with the user search)

cardoso:
@RocketChat/ios

Closes #1265

### DIFF
--- a/Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
+++ b/Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
@@ -15,7 +15,7 @@ extension LinkingObjects where Element == Subscription {
     }
 
     func filterBy(name: String) -> Results<Subscription> {
-        return self.filter("name CONTAINS %@", name)
+        return self.filter("name CONTAINS[c] %@", name)
     }
 }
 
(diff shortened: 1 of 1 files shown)


Rocket.Chat/Extensions/Models/SubscriptionExtensions.swift
```diff
@@ -25,6 +25,6 @@ extension Results where Element == Subscription {
     }
 
     func filterBy(name: String) -> Results<Subscription> {
-        return self.filter("name CONTAINS %@", name)
+        return self.filter("name CONTAINS[c] %@", name)
```
rafaelks: Do we have tests on this methods? I think would be something nice to have here if we don't. 👍

PR #1323, opened 2018-02-22
[FIX] Call super on viewDidLoad to load all tintColor settings

rafaelks:
@RocketChat/ios


cardoso:
What does this fix @rafaelks ?

rafaelks:
@cardoso It calls the super method from BaseNavigationController, that configures the navigationBar preferences all over the app.

PR #1332, opened 2018-02-22
[NEW] Add support to named colors, remove new lines on title from bots

rafaelks:
@RocketChat/ios

## Progress

- [x] Fixes
- [x] Tests


Rocket.Chat/Extensions/UIColorExtension.swift
```diff
@@ -19,6 +19,24 @@ extension UIColor {
         )
     }
 
+    // MARK: Color from strings (good|warning|danger)
+
+    static func normalizeColorFromString(string: String) -> UIColor {
+        if string == "warning" {
```
cardoso: What do you think of using a switch here instead?

Rocket.Chat/Extensions/StringExtensions.swift
```diff
@@ -96,7 +96,11 @@ extension String {
     }
 
     func removingWhitespaces() -> String {
-        return components(separatedBy: .whitespaces).joined()
+        return components(separatedBy: .whitespacesAndNewlines).joined()
```
cardoso: Should we update the tests for this class later?

Rocket.Chat/Extensions/UIColorExtension.swift
filipealva: I think that using a switch would be better too. Also, what do you guys think about having a `NamedColor` enum holding the color names as member values and the colors itself as properties?

Rocket.Chat/Extensions/StringExtensions.swift
```diff
@@ -96,7 +96,11 @@ extension String {
...
+    }
+
+    func removingNewLines() -> String {
+        return components(separatedBy: .newlines).joined()
```
cardoso: Can you add tests for this method in StringExtensionSpec.swift?

Example: string with spaces and newlines, but only newlines are removed

Rocket.Chat/Extensions/StringExtensions.swift
rafaelks: ✅

Rocket.Chat/Extensions/StringExtensions.swift
rafaelks: ✅

PR #1334, opened 2018-02-23
[FIX] Remove description indicator

### LINKED ISSUE #1287: [BUG] Attachment with no comment but reaction show »
The » indicator for the comment on the attachment is normally hidden, but when expanding the container to add reactions, it shows up, even if there is no comment. Will fix that somehow :)

soundstorm:
Remove description indicator when there is no description

Remove own typing indicator (again)

Closes #1287


rafaelks:
Looking good! Thank you @soundstorm!

PR #1343, opened 2018-02-24
[IMPROVEMENT] Make users searching in create new room (inviting user) case and diacritic insensitive

### LINKED ISSUE #1330: [Bug] Inviting users case sensitive
- Your Rocket.Chat app version: 2.0.2
- Your Rocket.Chat server version: 0.61.1
- Device (or Simulator) you're running with: SE

When creating a new channel, the selection for users is case sensitive.
How to test: 
- Create user Matheus (with upercase M)
- Create a new channel and add a user , starting with a lowercase "m"

=> You won't find it

artrmz:
@RocketChat/ios

Closes #1330

### DIFF
--- a/Rocket.Chat/Views/Form/Cells/MentionsTextFieldTableViewCell.swift
+++ b/Rocket.Chat/Views/Form/Cells/MentionsTextFieldTableViewCell.swift
@@ -70,7 +70,7 @@ class MentionsTextFieldTableViewCell: UITableViewCell, FormTableViewCellProtocol
         }
 
         let users = realm.objects(User.self).filter(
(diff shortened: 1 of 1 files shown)


Hrayr2112:
I guess you should use CONTAINS[cd] instead of CONTAINS[c]

artrmz:
Can add that if we want to be diacritic-insensitive as well?

Hrayr2112:
Ok, thanks)

Hrayr2112:
As I know its okay

rafaelks:
@artrmz Thank you!!! 👊

PR #1368, opened 2018-03-01
[NEW] Let user update name, username, email, password and avatar

### LINKED ISSUE #962: [NEW] Let user update name, username, email, password and avatar

filipealva:
@RocketChat/ios

Closes #962


Rocket.Chat/API/Clients/UploadClient.swift
```diff
@@ -14,8 +14,8 @@ struct UploadClient: APIClient {
         self.api = api
     }
 
-    func upload(roomId: String, data: Data, filename: String, mimetype: String, description: String, completion: (() -> Void)? = nil, versionFallback: (() -> Void)? = nil) {
-        let req = UploadRequest(
+    func uploadMessage(roomId: String, data: Data, filename: String, mimetype: String, description: String, completion: (() -> Void)? = nil, versionFallback: (() -> Void)? = nil) {
```
filipealva: @RocketChat/ios since this function is about uploading a media message to a room I thought it would be better to call it `uploadMessage` instead of just `upload`. Let me know if you guys think we should back to `upload`.

Rocket.Chat/Extensions/UIViewControllerExtension.swift
```diff
@@ -7,10 +7,59 @@
 //
 
 import UIKit
+import Photos
+import MobileCoreServices
 import ObjectiveC
 
 private var viewScrollViewAssociatedKey: UInt8 = 0
 
+protocol MediaPicker: AnyObject { }
```
filipealva: @RocketChat/ios here I would have to replicate the exact same code for launching the camera and photo library than we already had on `ChatViewController`, then I thought that it would be nice to add it into an extension. Firstly I thought it would be cool to put it into `UIViewController` extension, but it  doesn't make sense to have these media picking related code for all view controllers.
Assuming that we can't add constraints to `UIViewController` extensions I thought that the best way to emulate this behavior was to create a protocol (MediaPicker) and constrain it only to view controllers that implement the delegates needed to provide the media picking behavior.

It works for camera/photo library as for drawing, even though we only have the drawing code in one place at our codebase it will avoid us to replicate code in the future if we have to add drawing to another view controller.

To be honest I'm in doubt if this is an elegant solution or a tricky one 🤔 so please share your thoughts.

artrmz: I like it, feels really nice :+1:

Rocket.Chat/API/Clients/UploadClient.swift
rafaelks: Yes, much better now, specially since we have two upload methods :-)

Rocket.Chat/API/Requests/User/UpdateUserRequest.swift
```diff
@@ -0,0 +1,76 @@
...
+        self.user = user
+        self.password = password
+    }
+
+    func body() -> Data? {
+        guard let name = user.name,
+                let username = user.username,
+                let email = user.emails.first else {
+            return nil
+        }
```
rafaelks: We're using another style for guard... what do you think?

```swift
guard
    let name = user.name,
    let username = user.username,
    let email = user.emails.first
else {
    return nil
}
```

Rocket.Chat/Controllers/Settings/EditProfileTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+
+    @IBAction func saveProfile(_ sender: UIBarButtonItem) {
+        hideKeyboard()
+
+        guard let user = user, let userId = user.identifier else { return }
+        guard let name = name.text, let username = username.text, let email = email.text,
+            !name.isEmpty, !username.isEmpty, !email.isEmpty else {
+                Alert(key: "alert.update_profile_empty_fields").present()
+                return
+        }
```
rafaelks: guard style...

Rocket.Chat/Controllers/Settings/EditProfileTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            weakSelf.isUpdatingUser = false
+            if !weakSelf.isUploadingAvatar {
+                self?.stopLoading(sender: sender)
+            }
+        }
+
+        isUpdatingUser = true
+        let updateUserRequest = UpdateUserRequest(userId: userId, user: user, password: password)
+        api?.fetch(updateUserRequest, succeeded: { result in
+            stopUpdatingUser()
```
rafaelks: The call of this method here `stopUpdatingUser()` is because you want to stop the loaders? I think we can improve the name of this method to something like `stopLoading()` or something like that... what do you think?

Rocket.Chat/Extensions/UIViewControllerExtension.swift
rafaelks: @filipealva I think this is perfect for now, in case we need to have something more generic in the future we can change it. :-)

Rocket.Chat/Models/Mapping/UserModelMapping.swift
```diff
@@ -36,5 +36,37 @@ extension User: ModelMappeable {
...
+                    })
+                }
+
+                if realm.isInWriteTransaction {
+                    appendEmailsUsingRealmObjects()
+                } else {
+                    try? realm.write {
+                        appendEmailsUsingRealmObjects()
+                    }
+                }
```
rafaelks: Can't you use our extension of Realm here? Like this...

```swift
Realm.execute({ (realm) in
   ...
}
```

Rocket.Chat/API/Requests/User/UpdateUserRequest.swift
filipealva: It's really better. I'll change it.

Rocket.Chat/Resources/es.lproj/Localizable.strings
```diff
@@ -218,8 +227,10 @@
 "chat.messages.starred.list.title" = "Mensajes destacados";
 
 // Settings
+
```
rafaelks: Can you remove this new line? :D

Rocket.Chat/Extensions/UIViewControllerExtension.swift
cardoso: Cool @filipealva !

We could be doing this for our "Alerter" protocol also:

```swift
protocol Alerter: class {
    func alert(title: String, message: String, handler: ((UIAlertAction) -> Void)?)
}

extension Alerter where Self: UIViewController { // doing extension UIViewController: Alerter instead
    func alert(title: String, message: String, handler: ((UIAlertAction) -> Void)? = nil) {
        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
        alert.addAction(UIAlertAction(title: "OK", style: .default, handler: handler))
        present(alert, animated: true, completion: nil)
    }
}
```

But since almost all our view controllers need alerts, we just extend UIViewController.

Rocket.Chat/Controllers/Settings/EditProfileTableViewController.swift
filipealva: I think that `stopUpdatingUser()` was a bad naming idea, my bad. `stopLoading()` works!

Rocket.Chat/Models/Mapping/UserModelMapping.swift
filipealva: It seems that I can't...  When I try `Realm.execute` it throws a "wrong thread exception" and `Realm.executeOnMainThread` it says that it is already in a write transaction. 

but, going after it I could found a way to improve this code, take a look:

```swift
if let realm = realm {
    emails.forEach({ email in
        if let realmEmail = realm.object(ofType: Email.self, forPrimaryKey: email.identifier as AnyObject) {
            self.emails.append(realmEmail)
        }
    })
} else {
    self.emails.append(contentsOf: emails)
}
```

Rocket.Chat/Resources/es.lproj/Localizable.strings
filipealva: sure!

Rocket.Chat/Extensions/UIViewControllerExtension.swift
filipealva: Cool guys! thanks for commenting your thoughts :D

@cardoso that's true, in this case, it's better to just extend UIViewController.

rafaelks:
@filipealva By downloading your branch, I would suggest some UI changes, let me know what you think:

- Increase the font size of the text fields to 16pt (or 15pt maybe?)
- When I press "back" and I changed my information, alert the user that the changes will be discarted;
- Add a loader to the screen (maybe on the position of "save button") when loading information of the user;
- After saving the information, what do you think about showing a kind of overlay saying that the profile was updated? The experience today is only that the save button is back to the screen. 🤔 

It's looking great! 👍

filipealva:
@rafaelks I agree with you. Good points. I'm already working on them.

filipealva:
> After saving the information, what do you think about showing a kind of overlay saying that the profile was updated? The experience today is only that the save button is back to the screen.

@rafaelks What do you think about using `MBProgressHUD` to show something like this:

rafaelks:
@filipealva I like the way `MBProgressHUD` shows it. I would say for us to use it and in the future create our own custom view.

filipealva:
>@filipealva I like the way MBProgressHUD shows it. I would say for us to use it and in the future create our own custom view.

@rafaelks I agree with you, later we should create our own custom views.

filipealva:
@rafaelks All your feedbacks are implemented now. I just had to ignore this one:

> When I press "back" and I changed my information, alert the user that the changes will be discarded;

Because after implementing the UX flow suggested by @cardoso it is no longer necessary to notify the user since he taps on the "cancel" button to exit the `edit` mode and this by itself makes clear that he is discarding his unsaved changes.  

Now it is just missing to adjust for the new `update-user` endpoint.

rafaelks:
@filipealva New API is already merged and can be used in `open.rocket.chat` and `unstable.rocket.chat` already. Here is the PR: https://github.com/RocketChat/Rocket.Chat/pull/9719.

Don't forget the add the minimum requirement to `0.62.2` on the request model to update user's profile.

Rocket.Chat/Controllers/Settings/EditProfileTableViewController.swift
```diff
@@ -0,0 +1,437 @@
...
+//
+//  Created by Filipe Alvarenga on 27/02/18.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+import MBProgressHUD
+import SwiftyJSON
+
+class EditProfileTableViewController: UITableViewController, MediaPicker {
```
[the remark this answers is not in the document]
cardoso: Not sure if this warning is valid, but maybe try splitting the MediaPicker conformance into an extension below?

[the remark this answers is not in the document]
filipealva: The warning is valid, but the MediaPicker conformance is implemented into the MediaPicker extension, so it wouldn't decrease a line of the body :/

filipealva:
Guys, I tried to clean up the most I could without overthinking on how to fit the 400/200 lines to satisfy SwiftLint. 

Should we ignore the `type_body_length` and `file_length`?

PR #1372, opened 2018-03-03
[I18N] Update greek localization for new text

kmav:
@RocketChat/ios

This is the update of greek strings that failed before. 
I created a fix/kmav-greek-localisation branch from the related commit.
Only the greek Localisation.strings is different.


rafaelks:
@kmav Looking great! Thank you!

PR #1377, opened 2018-03-05
[NEW] Interactive keyboard dismissal added to ChatViewController / [FIX] Fixed the weird scrolling behavior and layout issues on chat rooms that one hasn't joined

### LINKED ISSUE #1378: [BUG] Layout is broken for rooms that user haven't joined on iPhone X
## General information

The UI is broken when user access a room that user haven't joined yet on iPhone X.

img_3945

Sameesunkaria:
@RocketChat/ios

out

- Interactive keyboard dismissal added to ChatViewController.
- Minor refactoring.

Since the Input Accessory View is tightly coupled with the SlackTextViewController, implementing the text bar as an Input Accessory View would require major modifications to the SlackTextViewController dependency. Until the dependency gets updated this is possibly the best way to get an interactive keyboard slide out.

- Fixed the weird scrolling behavior on chat rooms that one hasn't joined. 
- Fixed the layout issue of the chat preview view on devices with safe area insets. (iPhone X)

Closes #1378

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -28,17 +28,17 @@ final class ChatViewController: SLKTextViewController {
     @IBOutlet weak var buttonScrollToBottom: UIButton!
     var buttonScrollToBottomMarginConstraint: NSLayoutConstraint?
 
-    var showButtonScrollToBottom: Bool = false {
+    var scrollToBottomButtonIsVisible: Bool = false {
         didSet {
             self.buttonScrollToBottom.superview?.layoutIfNeeded()
 
-            if self.showButtonScrollToBottom {
-                self.buttonScrollToBottomMarginConstraint?.constant = -self.textInputbar.frame.height - 40
+            if self.scrollToBottomButtonIsVisible {
+                self.buttonScrollToBottomMarginConstraint?.constant = (self.textInputbar.frame.origin.y - self.view.frame.height) - 40
             } else {
                 self.buttonScrollToBottomMarginConstraint?.constant = 50
             }
 
-            if showButtonScrollToBottom != oldValue {
+            if scrollToBottomButtonIsVisible != oldValue {
                 UIView.animate(withDuration: 0.5) {
                     self.buttonScrollToBottom.superview?.layoutIfNeeded()
                 }
@@ -321,7 +321,7 @@ final class ChatViewController: SLKTextViewController {
         let sizeHeight = collectionView?.contentSize.height ?? 0
         let offset = CGPoint(x: 0, y: max(sizeHeight - boundsHeight, 0))
         collectionView?.setContentOffset(offset, animated: animated)
-        showButtonScrollToBottom = false
+        scrollToBottomButtonIsVisible = false
     }
 
     func resetMessageSending() {
@@ -1097,7 +1097,7 @@ extension ChatViewController: UICollectionViewDelegateFlowLayout {
 extension ChatViewController {
     override func scrollViewDidScroll(_ scrollView: UIScrollView) {
         super.scrollViewDidScroll(scrollView)
-        
+
         updateKeyboardConstraints()
 
         if scrollView.contentOffset.y < -10 {
@@ -1106,7 +1106,7 @@ extension ChatViewController {
             }
         }
 
-        showButtonScrollToBottom = !chatLogIsAtBottom()
+        scrollToBottomButtonIsVisible = !chatLogIsAtBottom()
     }
 }
 
@@ -1186,5 +1186,3 @@ extension ChatViewController {
         }
     }
 }
-
-
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -143,7 +143,7 @@ final class ChatViewController: SLKTextViewController {
     }
 
     deinit {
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -349,6 +341,82 @@ final class ChatViewController: SLKTextViewController {
...
+            keyboardHC = self.view.constraints.first {
+                ($0.firstItem as? UIView) == self.view &&
+                    ($0.secondItem as? SLKTextInputbar) == self.textInputbar
+            }
+        }
+
+        // Adding textInputBar background so that the app can support devices with safe area insets.
+        // The tool bar (textInputBar) background sometimes dissapears on keyboard slide outs,
+        // with no real fix for it provided by Apple in UIKit.
+        updateTextInputbarBackground()
```
rafaelks: 👍

Sameesunkaria:
@rafaelks Since even if we are checking for the location of the finger on the screen, we would need to calculate the `keyboardWindowHeight` and `keyboardYOriginInWindow` anyway. I believe checking for the finger coordinates would not be any faster (or possibly slower, since we are listening for gesture recognizer events simultaneously with the collection view) than just implementing the enhancements below.

The only 2 places that can help in improving the performance are:
- Not updating the constraint constant if they are the same. (Not sure if setting the constant to the same value as it was previously, would cause it to redraw. Any inputs maybe?)
- Making the old background for textInputView transparent once, currently it is being repeated on every consecutive call.

I'll make those changes.

Sameesunkaria:
@rafaelks I have made those changes.

rafaelks:
@Sameesunkaria Thank you! This feature also have bugs on channels that user haven't joined yet. To reproduce, try logging in `open.rocket.chat` and opening some room that you haven't joined, like `#iosnativeapp` or `#androidnativeapp` and then scroll to the bottom. You'll see some weird behaviours.

Sameesunkaria:
@rafaelks I get the weird scrolling behavior on the App Store release version of the app as well.

rafaelks:
@Sameesunkaria Oh, that's true! Good finding!

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -349,6 +341,90 @@ final class ChatViewController: SLKTextViewController {
         dataController.lastSeen = subscription?.lastSeen ?? Date()
     }
 
+    // MARK: Handeling Keyboard
+
+    weak var keyboardHC: NSLayoutConstraint?
+    weak var textInputbarHC: NSLayoutConstraint?
```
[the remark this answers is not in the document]
Sameesunkaria: Changed them to be a little more descriptive. Was following the naming convention used by the `SLKTextViewController`.

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -28,17 +28,17 @@ final class ChatViewController: SLKTextViewController {
     @IBOutlet weak var buttonScrollToBottom: UIButton!
     var buttonScrollToBottomMarginConstraint: NSLayoutConstraint?
 
-    var showButtonScrollToBottom: Bool = false {
+    var scrollToBottomButtonIsVisible: Bool = false {
```
cardoso: 👍

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -283,53 +285,43 @@ final class ChatViewController: SLKTextViewController {
...
-        autoCompletionView.register(UINib(
-            nibName: "AutocompleteCell",
-            bundle: Bundle.main
-        ), forCellReuseIdentifier: AutocompleteCell.identifier)
-
-        autoCompletionView.register(UINib(
-            nibName: "EmojiAutocompleteCell",
-            bundle: Bundle.main
-        ), forCellReuseIdentifier: EmojiAutocompleteCell.identifier)
+        typealias NibCellIndentifier = (nibName: String, cellIdentifier: String)
```
cardoso: Thanks!!

Rocket.Chat/Controllers/Chat/ChatViewController.swift
cardoso: 👍

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -878,9 +869,26 @@ final class ChatViewController: SLKTextViewController {
...
+
             view.addSubview(previewView)
+
+            NSLayoutConstraint.activate([
+                previewView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+                previewView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+                previewView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+                ])
+
+            collectionView?.bottomAnchor.constraint(equalTo: previewView.topAnchor).isActive = true
```
Sameesunkaria: Just for future reference, this constraint is completely valid and does not conflict with any of the constraints added by the `SLKTextViewController`. `SLKTextViewController` has added a height constraint to the `collectionView`, but since it's priority is `750`, the `bottomAnchor` constraint will always take precedence.

This constraint automatically gets removed when the previewView, leaves its superview. This allows for a normal operation when the `textInputbar` is visible and the `SLKTextViewController` takes over the control of the layout.

rafaelks:
@Sameesunkaria Amazing! Thank you very much! 👊

PR #1382, opened 2018-03-07
[IMPROVEMENT] Interactive keyboard dismissal improved

Sameesunkaria:
@RocketChat/ios

- The text input bar no longer lags 1 frame behind the keyboard.
- Should have better performance characteristics, since the new implementation is only called when the frame of the keyboard actually changes (unlike previously, where the method was called inside `scrollViewDidScroll`)

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -315,6 +315,10 @@ final class ChatViewController: SLKTextViewController {
         scrollToBottomButtonIsVisible = false
     }
 
+    internal func resetScrollToBottomButtonPosition() {
+        scrollToBottomButtonIsVisible = !chatLogIsAtBottom()
+    }
+
     func resetMessageSending() {
         textView.text = ""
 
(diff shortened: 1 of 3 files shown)


rafaelks:
@Sameesunkaria Thanks for the PR! 👍 

There's one issue with your changes, is that: if I'm scrolling thought the list of messages and then I press the textView to start typing a new message, the scroll goes to the bottom again. It shouldn't change the position if I'm looking into older messages. I understand why it's happening that, but can you do that only if user is at the bottom of the list of messages?

Sameesunkaria:
Oh, I believe the "scroll to bottom when keyboard is shown" behavior was already implemented previously. But I can see how it would get irritating cause now the keyboard slides out automatically. I'll get that fixed!

Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -459,10 +463,6 @@ final class ChatViewController: SLKTextViewController {
         }
     }
 
-    override func textViewDidBeginEditing(_ textView: UITextView) {
-        scrollToBottom(true)
-    }
-
```
Sameesunkaria: Was rather straight forward.

Sameesunkaria:
@rafaelks It should all work as expected now!

Sameesunkaria:
@rafaelks it would be nice if you could merge this with 2.1.0 beta.

rafaelks:
Looking good! Thank you!!!

PR #1384, opened 2018-03-07
[IMPROVEMENT] Improvements on the App Icon selection

### LINKED ISSUE #1285: [IMPROVEMENT] Change text displayed to user on changing icon
https://github.com/RocketChat/Rocket.Chat.iOS/blob/2b2707ed9059b6deb34b364204321d3ec30a4a9d/Rocket.Chat/Resources/en.lproj/Localizable.strings#L203

This text is quite long and not grammatically correct. I suggest changing the text to something like 'Choose an icon'

rafaelks:
@RocketChat/ios

Closes #1285


artrmz:
Looking great!

Rocket.Chat/Views/Cells/Preferences/ReusableViewText.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Rafael Kellermann Streit on 07/03/18.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+final class ReusableViewText: UICollectionReusableView {
+
+    static let identifier = "ReusableViewTextIdentifier"
```
filipealva: Nothing specifically related to this PR, just a note: what if we start using `String(describing: Self.self)` for identifiers like this one in the future?

rafaelks: Sounds good to me... could even be part of a Generic? 🤔

filipealva: Cool. Yes, it could :D

PR #1388, opened 2018-03-08
update the pod file

baquer:
@RocketChat/ios
Update the pod file for the compatibility of Xcode , as it shows an error while building the app
Also need to convert few codes in Swift 4

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,11 +1,11 @@
 PODS:
   - 1PasswordExtension (1.8.5)
-  - Crashlytics (3.9.3):
-    - Fabric (~> 1.7.2)
-  - Fabric (1.7.2)
+  - Crashlytics (3.10.1):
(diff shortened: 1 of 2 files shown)


artrmz:
What error are you getting? Albo please uparte your CocoaPods to the latest version since you use rc which is outdated.

rafaelks:
@baquer Can you describe what issue are you having? Not sure I understood your PR.

baquer:
Actually when i perform the Swift conversion to Swift 4 its gives an error in RCMarkdownParser .

artrmz:
RCMarkdownParser is a dependency defined in Podfile and it is configured to use Swift 3.1. If you want to update it to Swift 4 you should do it in this repo https://github.com/RocketChat/RCMarkdownParser and then update Podfile to configure it to use Swift 4 for RCMarkdownParser

PR #1379, opened 2018-03-05
[NEW] Implement Share Extension

cardoso:
@RocketChat/ios

Closes #820 

TODO:
- [x] Change servers screen title to small
- [x] Use localized description for HTTP errors
- [x] Remake file compose cell
- [x] Disable compose view when sending
- [x] Let user cancel sending
- [x] Let user retry if error happened

To reviewers:

- You can test sending urls, text, images, videos, other files (pdf)
- Test with an old and fresh install. (this feature performs a migration of both UserDefaults and Realm databases to a new folder)
- There is a lot of code to support sharing of multiple content, but we decided to block it for now while we decide how to do the UX for this.
- I had to make a lot of modifications to the main app's code to make it reusable in the extension. (mainly splitting extension into multiple files)
- This was made with a redux-like architecture.

PR #1395, opened 2018-03-10
[I18N] Translate missing polish strings

artrmz:
@RocketChat/ios

Can translate strings from #1368 and #1379 here as well, just let me know once they are merged.


rafaelks:
Thank you @artrmz!!!

PR #1402, opened 2018-03-13
[BUG] App does not render preview picture for videos directly 

### LINKED ISSUE #1268: [BUG] App does not render preview picture for videos directly
- Your Rocket.Chat app version: 1.8.0
- Your Rocket.Chat server version: 0.61.0
- Device (or Simulator) you're running with: iPhone 5S

**If you upload a video the preview picture is not rendered directly **
If a user uploads a video, the user will not get a preview picture. It does not matter how long the user waits for the picture. The upload animation will not disappear. 

img_0232

The user has to switch to another channel and switch back in order to get a preview of the uploaded video. 

I tested it by uploading a very short video (about 2-3 seconds long).

filipealva:
@RocketChat/ios

Closes #1268

### DIFF
--- a/Rocket.Chat/Models/Attachment.swift
+++ b/Rocket.Chat/Models/Attachment.swift
@@ -61,7 +61,7 @@ class Attachment: BaseModel {
     var videoThumbPath: URL? {
         let path = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
         let documents = path[0]
-        return documents.appendingPathComponent("\(identifier ?? "temp").png")
+        return documents.appendingPathComponent("\(!title.isEmpty ? title + type.rawValue : "temp").png")
     }
 
     var fields = List<AttachmentField>()
--- a/Rocket.Chat/Models/Attachment.swift
+++ b/Rocket.Chat/Models/Attachment.swift
@@ -61,7 +61,8 @@ class Attachment: BaseModel {
     var videoThumbPath: URL? {
         let path = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
         let documents = path[0]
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Models/Attachment.swift
```diff
@@ -61,7 +61,7 @@ class Attachment: BaseModel {
     var videoThumbPath: URL? {
         let path = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
         let documents = path[0]
-        return documents.appendingPathComponent("\(identifier ?? "temp").png")
```
filipealva: This is actually another bugfix.  Using this `identifier` to name the video thumb we were writing to disk but only being able to retrieve the cached thumb while the `Attachment` was in memory because every time we map an `Attachment` we set its identifier as a random string, so every time we load it on the chat we write a new file for the new random string identifier with the same data that should be cached. It was making a significant impact on our storage usage.

rafaelks: Good finding! :-)

Rocket.Chat/Views/Cells/Chat/ChatMessageVideoView.swift
```diff
@@ -68,20 +66,28 @@ final class ChatMessageVideoView: ChatMessageAttachmentView {
...
         }
     }
 
+    fileprivate func stopLoadingPreview(_ thumbnail: UIImage?) {
+        DispatchQueue.main.async {
+            self.activityIndicator.stopAnimating()
+            self.imageViewPreview.image = thumbnail
+            self.buttonPlay.isHidden = false
+        }
+    }
```
rafaelks: 👍

PR #1420, opened 2018-03-26
[FIX] FLEX cell in Preferences wasn't being selected

rafaelks:
@RocketChat/ios


Rocket.Chat/Controllers/Settings/PreferencesViewController.swift
```diff
@@ -16,6 +16,11 @@ import FLEX
 
 final class PreferencesViewController: UITableViewController {
 
+    private let kSectionProfile = 0
+    private let kSectionSettings = 1
+    private let kSectionInformation = 2
+    private let kSectionFlex = 3
+
```
filipealva: 👍

PR #1438, opened 2018-03-26
[CHORE] Reduce warnings

### LINKED ISSUE #1436: [CHORE] Resolve all project warnings

cardoso:
@RocketChat/ios

Closes #1436


Rocket.Chat/Controllers/Auth/AuthViewController+LoginServices.swift
```diff
@@ -0,0 +1,176 @@
+//
```
rafaelks: Following the ObjC categories style? 🤓

cardoso: What do you think? 🤔

rafaelks: @cardoso We're not using this approach on the other extensions... I like it, but then I think we would have to change the other files as well. 🙃

Rocket.Chat/API/Clients/MessagesClient.swift
```diff
@@ -58,7 +58,6 @@ struct MessagesClient: APIClient {
         }, errored: { error in
             switch error {
             case .version:
-                // TODO: Remove SendMessage Fallback + old methods after Rocket.Chat 1.0
```
cardoso: Let's just search for "case .version:" when we decide to remove fallbacks. That way we don't pollute with so many warnings.

cardoso:
Ready for review @rafaelks

Rocket.Chat/API/Clients/MessagesClient.swift
rafaelks: 👍

PR #1460, opened 2018-03-29
[NEW] Support pinning permissions & settings

### LINKED ISSUE #1429: [NEW] Support "Message_AllowPinning" setting

rafaelks:
@RocketChat/ios

Closes #1429

## Description

This PR implements the settings and permissions that allow (or not) the user to pin a message.

- [x] Setting: `Message_AllowPinning`
- [x] Permission: `pin-message`
- [x] Tests

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
@@ -73,13 +73,9 @@ extension ChatViewController {
             let pinMessage = message.pinned ? localized("chat.message.actions.unpin") : localized("chat.message.actions.pin")
             let pin = UIAlertAction(title: pinMessage, style: .default, handler: { (_) in
                 if message.pinned {
-                    MessageManager.unpin(message, completion: { (_) in
-                        // Do nothing
-                    })
+                    MessageManager.unpin(message)
                 } else {
-                    MessageManager.pin(message, completion: { (_) in
-                        // Do nothing
-                    })
+                    MessageManager.pin(message)
                 }
             })
 
@@ -128,7 +124,7 @@ extension ChatViewController {
                 return
             }
 
-            var _messageToResend: (identifier: String, text: String)? = nil
+            var messageToResend: (identifier: String, text: String)? = nil
 
             Realm.executeOnMainThread { realm in
                 guard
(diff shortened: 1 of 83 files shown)


Rocket.Chat/Models/AuthSettings.swift
```diff
@@ -81,6 +81,8 @@ final class AuthSettings: BaseModel {
     @objc dynamic var hideMessageUserRemoved: Bool = false
 
     // Message
+    @objc dynamic var messageAllowPinning = true
```
filipealva: @rafaelks Doesn't this naming pattern sound weird for you? 

"message allow pinning" <- Actually is not the message that allows or deny pinning, it is the server, right?

We could use `serverAllowsMessagePinning` and it would be more meaningful. Or we could even put it in the user perspective and name it `isAllowedToPinMessages` because even though is a server setting it is related to an action that the user performs. 

Does it make any sense?

Rocket.Chat/Controllers/Chat/ChatControllerMessageActions.swift
```diff
@@ -71,9 +67,26 @@ extension ChatViewController {
...
 
-        var actions = [react, pin, report, copy, reply, quote]
+        var actions = [react, reply, quote, copy, report]
+
+        if auth.canPinMessage(message) == .allowed {
+            let pinMessage = message.pinned ? localized("chat.message.actions.unpin") : localized("chat.message.actions.pin")
+            let pin = UIAlertAction(title: pinMessage, style: .default, handler: { (_) in
+                if message.pinned {
+                    MessageManager.unpin(message, completion: { (_) in
+                        // Do nothing
```
filipealva: @rafaelks Couldn't we make `completion` optional and make its default value `nil`? This way we avoid empty closures like this one.

Rocket.Chat/Models/AuthSettings.swift
cardoso: @filipealva I think we should follow the back-end naming "Message_AllowPinning"

filipealva: @cardoso Got it! 

I personally disagree with this, I rather follow the naming convention of each platform, also I think that part of the job of "mapping" it to our `.swift` model is to make it conform to our naming conventions. 

But if that's the pattern that we are adopting, no problems, but in this case I would say that using literally `Message_AllowPinning`  or `message_allowPinning` would be more meaningful for the developers to read.

cardoso: I agree that it's not the best naming. But It's not a big problem if we are consistent.

Probably the best way to do this would be to separate each namespace into a different struct with their own mapping functions. So we would access it like this:

```swift
settings.message.allowPinning
```

PR #1464, opened 2018-03-29
[FIX] Addresses prolonged freezes in some chat rooms

Sameesunkaria:
@RocketChat/ios

- The regex for emoji replacement is now fixed and does not go into a state of catastrophic backtracking.
- An edge case fixed where the emojis would not get replaced.
- Also, addresses the concerns raised by @cardoso in #1451 
- Added test cases to cover a variety of edge cases.

### DIFF
--- a/Rocket.Chat/External/RCEmojiKit/Emojione+Transform.swift
+++ b/Rocket.Chat/External/RCEmojiKit/Emojione+Transform.swift
@@ -9,12 +9,12 @@
 import Foundation
 
 extension Emojione {
-    static func transform(string: String) -> String {
+    static func transrform(string: String) -> String {
         var _string = string as NSString
 
-        let regex = try? NSRegularExpression(pattern: ":(\\w|-|\\+)+:", options: [])
+        let regex = try? NSRegularExpression(pattern: Emojione.regex, options: [])
         let ranges = regex?.matches(
-            in: _string as String,
+            in: string,
             options: [],
             range: NSRange(location: 0, length: string.count)
         ).map {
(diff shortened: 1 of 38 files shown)


Rocket.Chat/External/RCEmojiKit/Emojione+Transform.swift
```diff
@@ -9,7 +9,7 @@
 import Foundation
 
 extension Emojione {
-    static func transform(string: String) -> String {
+    static func transrform(string: String) -> String {
```
rafaelks: Ups, typo?

Sameesunkaria: Oh I meant to remove that function

Rocket.Chat/External/RCEmojiKit/NSAttributedString+CustomEmojis.swift
```diff
@@ -10,51 +10,65 @@ import Foundation
...
+        let matchRanges = ranges.map { range in regex?.matches(in: string.string, options: [], range: range).map { $0.range(at: 0) } ?? [] }
+        return matchRanges.reduce(into: [NSRange]()) { $0.append(contentsOf: $1) }
     }
 }
 
 extension String {
     func codeRanges() -> [NSRange] {
         let codeRegex = try? NSRegularExpression(pattern: "(```)(?:[a-zA-Z]+)?((?:.|\r|\n)*?)(```)", options: [.anchorsMatchLines])
-        let codeMatches = codeRegex?.matches(in: self, options: [], range: NSRange(location: 0, length: count)) ?? []
+        let codeMatches = codeRegex?.matches(in: self, options: [], range: NSRange(location: 0, length: utf16.count)) ?? []
```
rafaelks: 😅

PR #1400, opened 2018-03-12
Added no internet connection functionality , As the internet gets dis…

baquer:
…connected it will pop up the message and when the network available it automatically gets connected

@RocketChat/ios
screen shot 2018-03-13 at 1 20 02 am

PR #1465, opened 2018-03-30
[NEW] Handle "No Internet Connection" state.

baquer:
@RocketChat/ios
closes #1400  and split the PR for two new features as suggested by @rafaelks


baquer:
screen shot 2018-03-30 at 4 47 12 pm

rafaelks:
@baquer Thanks for the PR! Why do you think we need to use Reachability in order to alert the user? I think an error on the all of the API is enough. Have you tried that already?

rafaelks:
Closing this one duo to inactivity.

PR #1415, opened 2018-03-23
[FIX] Invalid url path building

filipealva:
@RocketChat/ios

Closes #1414

PR #1277, opened 2018-02-16
[FIX] Enable support for Rocket hosted on subdirectory

jpowie01:
@RocketChat/ios

Closes #900

These changes make sure that API Requests will follow HTTP protocol with applied subdirectory on which Rocket.Chat Server is hosted. Also, saves HTTP connection to the database instead of WebSocket one.

**Note:** It would be nice if anyone would like to check it out and even countinue work on this as I'm not familiar with Swift/iOS/Rocket.Chat.iOS at all. It's my proposal for changes :)

PR #1470, opened 2018-04-01
[FIX] Build URL that may use subdirectory

jpowie01:
@RocketChat/ios

A few weeks ago I've posted a change that introduced support for Rocket.Chat hosted on subdirectory (https://github.com/RocketChat/Rocket.Chat.iOS/pull/1277). Next day, @filipealva posted a fix that actually broken my patch but unblock him due to slash at the end of the `api.host` (https://github.com/RocketChat/Rocket.Chat.iOS/pull/1415). It resulted in a whole path replacement instead of applying requested path to the end of the host address!

There are four cases where Rocket.Chat could be hosted and where `/api/v1/me` should point to:
- `https://instance.com` -> `https://instance.com/api/v1/me`
- `https://instance.com/` -> `https://instance.com/api/v1/me`
- `https://instance.com/rocket` -> `https://instance.com/rocket/api/v1/me`
- `https://instance.com/rocket/` -> `https://instance.com/rocket/api/v1/me`

That's why this patch changes `=` (and even my previous `+=`) to `appendingPathComponent` method, which should work fine in all above cases. I've tested it on my own instance hosted on subdirectory and `https://open.rocket.chat`.

### DIFF
--- a/Rocket.Chat/API/APIRequest.swift
+++ b/Rocket.Chat/API/APIRequest.swift
@@ -57,8 +57,8 @@ extension APIRequest {
     }
 
     func request(for api: API, options: APIRequestOptions = .none) -> URLRequest? {
-        var components = URLComponents(url: api.host, resolvingAgainstBaseURL: false)
-        components?.path = path
+        let requestedUrl = api.host.appendingPathComponent(path)
+        var components = URLComponents(url: requestedUrl, resolvingAgainstBaseURL: false)
         components?.query = query
 
         if let optionsQuery = options.query {
--- a/Rocket.Chat/API/APIRequest.swift
+++ b/Rocket.Chat/API/APIRequest.swift
@@ -57,8 +57,8 @@ extension APIRequest {
     }
 
     func request(for api: API, options: APIRequestOptions = .none) -> URLRequest? {
-        let requestedUrl = api.host.appendingPathComponent(path)
-        var components = URLComponents(url: requestedUrl, resolvingAgainstBaseURL: false)
+        var components = URLComponents(url: api.host, resolvingAgainstBaseURL: false)
+        components?.path += path
         components?.query = query
 
         if let optionsQuery = options.query {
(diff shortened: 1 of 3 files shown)


filipealva:
Hi @jpowie01,

Sorry about that, I should've tested if your feature stayed working after I reverted your modifications on url path building, but my main intention was to revert the bug it have introduced that broke requests for instances such as `https://open.rocket.chat/`. 

Now I'm testing your changes manually and I'm seeing the same behavior that it was causing before I reverted it on #1415.  Look at the screen shot below:

screen shot 2018-04-01 at 16 32 40

It's not happening on current `develop` due to my fix but it's happening on your branch. By using `appendingPathComponent` or concatenating (`+=`) we have a double slash on the path making the requests to be invalid. This is why I pushed a fix returning it to `=` on #1415. 

This is an important feature, we want to have it and we deeply appreciate your efforts to make it, but we need to find another way to implement it because in this way the app will break for the majority of the servers that use our solution.

filipealva:
The code below could be a way to fix this double slash issue: 



```diff 
#### URLExtension.swift ####

    func httpServerURL() -> URL? {
        var components = URLComponents()
        components.scheme = "https"
        components.host = self.host != nil ? self.host : self.path
        components.path = self.host != nil ? self.path : ""
        components.port = self.port != nil ? self.port : nil

        var newURL = components.url
        if components.path.contains("websocket") {
            newURL = newURL?.deletingLastPathComponent()
        }

+       if var newURLString = newURL?.absoluteString, newURLString.last == "/" {
+          newURLString.removeLast()
+          newURL = URL(string: newURLString)
+      }

        return newURL
    }
```

@rafaelks @cardoso What do you guys think about it?

jpowie01:
Sure, I’m open for any suggestion :) Your proposal looks good to me!

rafaelks:
@filipealva @jpowie01 Looks good to me! :v:

filipealva:
@rafaelks @jpowie01 great!

@jpowie01 could you please implement these changes so we can review and merge this pull request instead of opening another one?

jpowie01:
Yeah, sure! I've posted a change :) It would be nice if you could test it.

rafaelks:
@jpowie01 Your tests aren't passing... can you take a look please? :-)

jpowie01:
Sure, but I’ll take a look in a few hours. Don’t have much time recently ;)

rafaelks:
Looking good! 👍

PR #1476, opened 2018-04-02
[IMPROVEMENT] Migrate "Public Settings" from RealTime to REST

### LINKED ISSUE #1433: [IMPROVEMENT] Migrate "Public Settings" from RealTime to REST
Requires server version **0.62.2**. Fallback to RealTime.
API documentation: https://rocket.chat/docs/developer-guides/rest-api/settings/public/.

filipealva:
@RocketChat/ios

Closes #1433


Rocket.Chat/Controllers/Auth/AuthViewControllerLoginServices.swift
```diff
@@ -50,12 +50,6 @@ extension AuthViewController {
     }
 
     func setupLoginServices() {
-        self.loginServicesToken?.invalidate()
-
-        self.loginServicesToken = LoginServiceManager.observe { [weak self] changes in
-            self?.updateLoginServices(changes: changes)
-        }
```
cardoso: If we do this, we break Custom OAuth methods since they are not contained in public settings. Can you undo these changes?

You can confirm this by connecting to my server: cardoso.rocket.chat

It has two custom OAuth methods: GITHUB and OPEN

filipealva: @cardoso I looked carefully at this when removing the code. We had a redundancy in the way we were adding the Custom OAuth buttons to the screen that just doesn't make any sense.

If the API works correctly it will add the Custom OAuth buttons to the screen just as it used to do with this duplicated logic. The issue here is that `public settings` request for `cardoso.rocket.chat` is returning `false` for GitHub OAuth:

```
     "type" : "boolean",
      "value" : false,
      "_id" : "Accounts_OAuth_Github"
``` 

I updated the code to rely on `public settings` REST request because it doesn't make sense to have these redundancy ways to handle the rendering of Custom OAuth buttons. It's weird because if you test with `open.rocket.chat` it works, the `Accounts_OAuth_Github` (and others) returns true and its button is rendered on the screen. 

Now I'm not sure if we should rely on `public settings` REST request at all. Because if it can behave this inconsistently at the point of returning `false` for a field that should be `true` maybe it's not stable enough and can cause this kind of issue for other fields too.

Rocket.Chat/Managers/Model/AuthSettingsManager.swift
```diff
@@ -26,20 +26,64 @@ final class AuthSettingsManager {
         }
     }
 
-    static func updatePublicSettings(_ auth: Auth?, completion: @escaping MessageCompletionObject<AuthSettings>) {
+    static func updatePublicSettings(serverVersion: Version? = nil, apiHost: URL? = nil, _ auth: Auth?, completion: @escaping MessageCompletionObject<AuthSettings>) {
+        var api: API?
```
cardoso: This can be `let`

Rocket.Chat/Managers/Model/AuthSettingsManager.swift
```diff
@@ -26,20 +26,64 @@ final class AuthSettingsManager {
...
+            }
+        })
+    }
+
+    private static func websocketUpdatePublicSettings(_ realm: Realm?, _ auth: Auth?, completion: @escaping MessageCompletionObject<AuthSettings>) {
         let object = [
             "msg": "method",
             "method": "public-settings/get"
-        ] as [String: Any]
+            ] as [String: Any]
```
cardoso: Can you undo this change?

Rocket.Chat/API/Requests/General/PublicSettingsRequest.swift
```diff
@@ -0,0 +1,86 @@
...
+                "Message_HideType_au",
+                "Message_HideType_mute_unmute",
+                "Message_HideType_ru",
+                "Message_ShowDeletedStatus",
+                "Message_AllowDeleting",
+                "Message_AllowDeleting_BlockDeleteInMinutes",
+                "Message_ShowEditedStatus",
+                "Message_AllowEditing",
+                "Message_AllowEditing_BlockEditInMinutes",
+                "Accounts_CustomFields"
```
cardoso: I think these would be more maintainable if they were in a separate array then you could interpolate them into the query.

```swift
    """
    fields={
        "type":1
    }&query={
        "_id":{
            "$in":[
                \(settings.map { "\"\($0)\"" }.joined(separator: ","))
            ]
        }
    }
"""
```

settings could be stored in a static property of AuthSettings then passed in the request's init

filipealva: @cardoso I think that it would still be the same pain to maintain as it already is if we split a huge hardcoded string into an array with a lot of hardcoded strings. 

The pros of your suggestion in my opinion:

- We at least split the fields from the URL, so we make the string smaller, but in different places. 
- It gives us the ability to easily exclude some fields from the array if we wind up needing to request just one setting.

The cons: 

- It adds a new property that in the current scenario will never be changed because we always request all the fields, so it would be just a constant array of hardcoded strings. 

In my opinion, we will add complexity to our code and gain almost nothing thinking about maintainability. I mean, we would not have the giant string in our code, but we would wind up adding more code only to break this string in a way that it would look better but honestly, I don't think it will be more maintainable because in some way we will need to have this huge string. 

If it was not a hardcoded string that will never change, but a URL built with dynamic parameters I would 100% agree with you. 

Besides respectfully disagreeing with your point about maintainability I have nothing against your suggestion, I'm just answering what I think :P 

Let me know if you still want me to implement it, so I will do it.

Rocket.Chat/Managers/Model/AuthSettingsManager.swift
filipealva: Sure 👌

filipealva:
@cardoso Great! Just fixed the conflitcs

PR #1480, opened 2018-04-03
[NEW] Allow user to see just the mentions of some channel

### LINKED ISSUE #344: [NEW] Allow user to see just the mentions of some channel
## General information

The user needs to have another option in our Channel Info screen that will let him see all the messages with mentions to him in a list.

To implement this list, there's already a list of messages being used to display the pinned and starred messages.

### How to request only the mentions

(to be documented).

filipealva:
@RocketChat/ios

I would like to provide some disclosure on why I had to make `MessagesListViewData` work with two different requests: 

This feature is about listing messages that contain mentions about the current user. There's a view controller that handles the same behavior for pinned and starred messages. I thought that it would be nice to use the same view controller since the behavior is all the same, the only thing that should change is a parameter for querying the messages, and it would be as simple as that but we haven't followed the standard we used for querying pinned and starred messages on the backend side. 

Because of that, I had to create an almost duplicated `APIRequest` in the iOS side as well and also make `MessagesListViewData` to conform with this new almost duplicated request, which adds a little bit of complexity and kind of duplicated logic to do the same thing in the end.

I'm registering it here for you guys to keep it in mind while reviewing the changes and as a reminder of things we should not do in our future API (`v2`).

Closes #344


Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
```diff
@@ -10,7 +10,7 @@ import UIKit
 import RealmSwift
 
 extension APIResult where T == SubscriptionMessagesRequest {
-    func getMessages() -> [Message?]? {
+    func fetchMessagesFromRealm() -> [Message]? {
```
rafaelks: What you mean with `fetchMessagesFromRealm()` here? This method isn't getting the values from JSON?

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
```diff
@@ -54,41 +65,62 @@ class MessagesListViewData {
...
+                        showing: result.count,
+                        total: result.total,
+                        completion: completion
+                    )
+                }, errored: { _ in
+                    Alert.defaultError.present()
+                })
+            } else {
+                let request = SubscriptionMessagesRequest(roomId: subscription.rid, type: subscription.type, query: query)
+                API.current()?.fetch(request, options: options, succeeded: { result in
```
rafaelks: Looks like the call of the `fetch()` could be the same for both requests?

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
filipealva: @rafaelks It looks a bit confusing, but there's a difference between `fetchMessagesFromRealm` and the `messages` property of the result. The difference is: `messages` just map the values from JSON without using Realm to fetch subobjects that only come with the `primary key` from the server. 

This is why we have this method in the first place, it was there before I started working on it, I just decided to change its name to make it clear that it is not just mapping the JSON, but actually fetching the objects referenced in JSON from Realm. 

In practice, the difference consists in passing the current Realm as a parameter in `Message.map` instead of `nil`.

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
filipealva: @rafaelks Since we have two different requests but they're both `APIRequest` you probably are thinking that we could just initialize a generic `var request: APIRequest` with the generic type based on the condition, but we can't use it to call `API.fetch` because it would not be able to identify which generic type we would be calling and it would throw a compilation error saying that the `API.fetch` call is ambiguos.

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
rafaelks: 👍

Rocket.Chat/Controllers/Chat/MessagesListViewController.swift
rafaelks: 👍

PR #1500, opened 2018-04-05
[HOTFIX] Requests throwing version error mistakenly

### LINKED ISSUE #1499: [BUG] Requests throwing version error mistakenly
- Your Rocket.Chat app version: 2.1.1
- Your Rocket.Chat server version: 0.63.0
- Device (or Simulator) you're running with: iPhone 6s

All requests with a required version with the `patch` greater than the `patch` of the current server version are being considered as greater than the current server version, no matter if the major or minor, i.e with the current logic `0.62.2` would be considered greater than `0.63.0`.

filipealva:
@RocketChat/ios

Closes #1499

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ DerivedData
 *.perspectivev3
 !default.perspectivev3
 xcuserdata
+xcshareddata
 
 ## Other
 *.xccheckout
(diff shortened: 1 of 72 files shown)


filipealva:
@rafaelks no problems. This fix is already done in #1476, so we don't need a new PR.

PR #1523, opened 2018-04-12
[NEW] Allow user to see all attachments in a channel/group/DM

### LINKED ISSUE #1314: [NEW] Allow user to see all attachments in a channel/group/DM

filipealva:
@RocketChat/ios

Closes #1314


rafaelks:
@filipealva What do you think about adding a very light (gray) border color to the thumb, and corner radius a 4? The first image is impossible to see right now, because it's most white.

simulator screen shot - iphone se - 2018-05-02 at 17 50 51

Rocket.Chat/API/Requests/Subscription/SubscriptionFilesRequest.swift
```diff
@@ -0,0 +1,68 @@
...
+        self.roomId = roomId
+        self.query = "roomId=\(roomId)&sort={\"uploadedAt\":-1}"
+    }
+}
+
+final class SubscriptionFilesResource: APIResource {
+    var files: [File]? {
+        return raw?["files"].arrayValue.map {
+            let file = File()
+            file.map($0, realm: Realm.current)
```
rafaelks: Save Realm instance before the loop?

Rocket.Chat/Controllers/Subscriptions/FilesListViewController.swift
```diff
@@ -0,0 +1,319 @@
...
+
+        // Checks if we do have the file in the system, before downloading it
+        if DownloadManager.fileExists(localFileURL) {
+            open()
+        } else {
+            let message = String(format: localized("chat.download.downloading_file"), file.name)
+            let loadingHUD = MBProgressHUD.showAdded(to: view, animated: true)
+            loadingHUD.label.text = message
+            // Download file and cache it to be used later
+            DownloadManager.download(url: fileURL, to: localFileURL) {
```
rafaelks: [weak self]?

Rocket.Chat/Controllers/Subscriptions/FilesListViewController.swift
```diff
@@ -0,0 +1,319 @@
...
+            }
+
+            DispatchQueue.main.async {
+                MBProgressHUD.hide(for: self.view, animated: true)
+                self.present(ImageViewerController(configuration: configuration), animated: false)
+            }
+        }
+
+        MBProgressHUD.showAdded(to: view, animated: true)
+        SDWebImageDownloader.shared().downloadImage(with: fileURL, options: [.useNSURLCache], progress: nil, completed: { _, data, _, _ in
```
rafaelks: [weak self]?

Rocket.Chat/Models/File.swift
```diff
@@ -0,0 +1,91 @@
...
+//  Rocket.Chat
+//
+//  Created by Filipe Alvarenga on 11/04/18.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+class File: BaseModel {
```
rafaelks: I imagine we can start using File in the future and remove the Attachment? 🤔

filipealva: @rafaelks it will depend on the API. I've created a different object for "File" because it seems that for the API they're two different things. The point is that many Attachments are basically files but not all Attachments can be represented by a File (text attachments, for example).

I think this is something that we will have to rethink together with the backend team.

Rocket.Chat/Controllers/Subscriptions/FilesListViewController.swift
filipealva: Honestly I don't know how it behave when we have a reference for `self` inside a sub function that is called inside an async-operation callback 🤔 I will update this sub function to a closure then capture the weak reference of self to make sure it will not cause any problem.

Rocket.Chat/Controllers/Subscriptions/FilesListViewController.swift
filipealva: Honestly I don't know how it behave when we have a reference for `self` inside a sub function that is called inside an async-operation callback 🤔 I will update this sub function to a closure then capture the weak reference of self to make sure it will not cause any problem.

rafaelks:
Looking very good! 👍

Rocket.Chat/Models/File.swift
```diff
@@ -0,0 +1,91 @@
...
+    var isVideo: Bool {
+        let videoTypes = ["mp4", "mov", "video", "flv"]
+        return videoTypes.compactMap { type.range(of: $0, options: .caseInsensitive) != nil ? true : nil }.first ?? false
+    }
+
+    var isAudio: Bool {
+        let audioTypes = ["mp3", "wav", "3gp", "audio"]
+        return audioTypes.compactMap { type.range(of: $0, options: .caseInsensitive) != nil ? true : nil }.first ?? false
+    }
+
```
mrqbmrqb: What if the file is audio mp4?
why not just `return type.range(of: "audio", options: .caseInsensitive) != nil` ?

The same thing for video and image.

rafaelks: Makes sense! 👍

rafaelks: @mrqbmrqb Made the changes here: https://github.com/RocketChat/Rocket.Chat.iOS/pull/1663.

PR #1528, opened 2018-04-12
[IMPROVEMENT] Share extension: support 2 attachment combos

### LINKED ISSUE #1525: [IMPROVEMENT] Share Extension: needs to support different kinds of attachments at the same time
We need to support this kind of configuration for now:

- [ ] 1 Text + 1 URL
- [ ] 1 Text + 1 Image
- [ ] 1 Text + 1 Video
- [ ] 1 URL + 1 Image
- [ ] 1 URL + 1 Video

cardoso:
@RocketChat/ios

Closes #1525

### DIFF
--- a/Rocket.Chat.ShareExtension/Info.plist
+++ b/Rocket.Chat.ShareExtension/Info.plist
@@ -35,7 +35,7 @@
 			<key>NSExtensionActivationRule</key>
 			<dict>
 				<key>NSExtensionActivationSupportsAttachmentsWithMaxCount</key>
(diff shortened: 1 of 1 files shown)


cardoso:
@rafaelks Medium requires us to accept 2 attachments, but it sends only one 🤔 

```swift
(lldb) po itemProviders
▿ 1 element
  - 0 : <NSItemProvider: 0x1c40d7ed0> {types = (
    "public.url"
)}
```

Do you have any other example? If not, I think we can just merge this.

PR #1531, opened 2018-04-13
[FIX] Not opening corresponding room after pressing notification

cardoso:
@RocketChat/ios

Looks like we had a regression after we started supporting paths 🤔


Rocket.Chat.xcodeproj/project.pbxproj
```diff
@@ -3317,21 +3300,6 @@
...
-			inputPaths = (
-			);
-			name = "📦 Copy Pods Resources";
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-Rocket.Chat/Pods-Rocket.Chat-resources.sh\"\n";
-			showEnvVarsInLog = 0;
-		};
```
rafaelks: 👀

cardoso: 🤔

PR #1540, opened 2018-04-14
[CHORE] Remove + from the name, increment version & fix Instabug for Release builds

rafaelks:
@RocketChat/ios


cardoso:
@rafaelks there are also some Rocket.Chat+ mentions in Localizable.strings files.

PR #1553, opened 2018-04-17
[IMPROVEMENT] Investigate and update models that unnecessarily inherit from BaseModel

### LINKED ISSUE #1474: [IMPROVEMENT] Investigate and update models that unnecessarily inherit from BaseModel
We usually make all models inherit from `BaseModel` but it requires the model to have a `primary key` and not always we have one, and when we don't have a primary key but we still inherit from `BaseModel` it forces us to generate a random primary key or even keep it nil which can lead to persisting data issues or even duplicated data on our local databases.

One example where it happens: `Mention.swift`.

filipealva:
@RocketChat/ios

Closes #1474

### DIFF
--- a/Rocket.Chat/Models/Attachment.swift
+++ b/Rocket.Chat/Models/Attachment.swift
@@ -10,13 +10,13 @@ import Foundation
 import SwiftyJSON
 import RealmSwift
 
-class AttachmentField: BaseModel {
+class AttachmentField: Object {
     @objc dynamic var short: Bool = false
     @objc dynamic var title: String = ""
     @objc dynamic var value: String = ""
 }
 
-class Attachment: BaseModel {
+class Attachment: Object {
     var type: MessageType {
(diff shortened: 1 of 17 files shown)


Rocket.Chat/Models/Attachment.swift
```diff
@@ -37,6 +37,7 @@ class Attachment: BaseModel {
         return titleLinkDownload && !titleLink.isEmpty
     }
 
+    @objc dynamic var identifier: String?
```
rafaelks: If we have identifier here, why removing it as a `BaseModel`?

filipealva: We need it to do at least the in-memory cache until we close #1556, but it doesn't need to be indexed as a primary key this is why it's not a `BaseModel`

rafaelks: Got it! 👍

PR #1555, opened 2018-04-17
[IMPROVEMENT] Clear chat when logging out

### LINKED ISSUE #1554: [IMPROVEMENT] Clear chat when logging out
One step of the logout process is to delete all objects on the current server database. The problem is: we do it with the chat view active and referencing a lot of messages that are being deleted. It may be causing an intermittent crash that happens when we hold active references to objects that are already deleted on the local database.

filipealva:
@RocketChat/ios

Closes #1554


rafaelks:
@filipealva Make sense the change, thank you! 👍

PR #1569, opened 2018-04-19
[IMPROVEMENT] Holding core chat references when adding a new server

### LINKED ISSUE #1568: [IMPROVEMENT] Holding core chat references when adding a new server
Since `SocketManager.connectionHandlers` is a dictionary it holds strong references to the values it stores. So when we go to add a new server we still hold in memory the core chat view controllers references on `SocketManager.connectionHandlers` when they should be gone and they keep interacting with the socket (reconnecting when we disconnect, sending commands, etc) which may cause problems when connecting to a new server not to say about performance issues due to heavy unnecessary instances being held in memory.

filipealva:
@RocketChat/ios

Since we already had problems caused by holding strong references on `connectionHandlers` I decided to refactor `connectionHandlers` from `Dictionary` to `NSMapTable` which is able to store weak references. 

Although `NSMapTable` lets the objects to be released from memory automatically it doesn't happen as fast as if we manually clear the references, so keep in mind that if you need to deallocate the references stored on `connectionHandlers` immediately to do a heavy operation (such as switching to another server) you better clear it manually to avoid performance issues older devices. 

Closes #1568

### DIFF
--- a/Rocket.Chat/Managers/AppManager.swift
+++ b/Rocket.Chat/Managers/AppManager.swift
@@ -99,6 +99,7 @@ extension AppManager {
     }
 
     static func reloadApp() {
+        SocketManager.sharedInstance.connectionHandlers.removeAllObjects()
         SocketManager.disconnect { (_, _) in
             DispatchQueue.main.async {
                 if AuthManager.isAuthenticated() != nil {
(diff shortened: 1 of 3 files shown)


rafaelks:
Clever solution @filipealva! Looking great! 👍

PR #1580, opened 2018-04-22
[CHORE] Update RCMarkdownParser (fix alt link parsing)

cardoso:
@RocketChat/ios

`<https://a.b|a<b>>`

Before:
before
After:
after

This may be related to a crash we have currently. Let's see if this fixes it 🙂


rafaelks:
Thank you @cardoso!!!

PR #1584, opened 2018-04-23
[NEW] Fabric tracking as an option

### LINKED ISSUE #1477: [NEW] Make the "tracking" optional in Settings
All kind of tracking in our app needs to be a user's choice. By default this flag is TRUE now, because we want to track the crashes, but if someone wants to disable it, the user needs to be able to disable via settings.

rafaelks:
@RocketChat/ios

Closes #1477


Rocket.Chat/Managers/Launcher/BugTrackingCoordinator.swift
```diff
@@ -14,8 +14,29 @@ import Crashlytics
...
+
+    static var isCrashReportingDisabled: Bool {
+        return UserDefaults.standard.bool(forKey: kCrashReportingDisabledKey)
+    }
+
+    static func toggleCrashReporting(disabled: Bool) {
+        UserDefaults.standard.set(disabled, forKey: kCrashReportingDisabledKey)
+
+        if disabled {
+            anonymizeCrashReports()
```
filipealva: @rafaelks if for `disabled` we just anonymize the crash reports, why not always launch on `prepareToLaunch` and just anonymize/identify on toggling?

PR #1590, opened 2018-04-24
Add Support for the WatchOS

tusharjindal353:
@RocketChat/ios

We should also develop this app for watch os for apple watches so that user can use the rich features even on the apple watches


rafaelks:
@tusharjindal353 What you mean with this pull-request?

Issue #1319, opened 2018-02-22
[NEW] Allow user with permissions to remove users from a room


rafaelks:
- [ ] Allow user to remove an user from a channel/group [(API)](https://rocket.chat/docs/developer-guides/rest-api/channels/kick/)

PR #1605, opened 2018-04-26
[IMPROVEMENT] Show action sheet when tapping usernames, mentions, etc

cardoso:
@RocketChat/ios

- Instead of just opening the direct message without prompting the user, now we show an action sheet.
- Also fixes real name/username not displaying according server settings in member cells. (used in members list and reactions)

Preview images:

iPhone:

### DIFF
--- a/Rocket.Chat/Views/ActionSheets/UserActionSheetPresenter.swift
+++ b/Rocket.Chat/Views/ActionSheets/UserActionSheetPresenter.swift
@@ -1,0 +1,34 @@
+//
+//  UserActionSheetPresenter.swift
+//  Rocket.Chat
+//
+//  Created by Matheus Cardoso on 4/26/18.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import Foundation
+
+protocol UserActionSheetPresenter {
+    func presentActionSheetForUser(_ user: User, source: (view: UIView?, rect: CGRect?)?)
+}
+
+extension UserActionSheetPresenter where Self: UIViewController {
+    func presentActionSheetForUser(_ user: User, source: (view: UIView?, rect: CGRect?)? = nil) {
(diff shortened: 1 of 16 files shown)


Rocket.Chat/Views/Cells/Chat/Info/MemberCell.swift
```diff
@@ -13,7 +13,7 @@ struct MemberCellData {
 
     var nameText: String {
         let utcText = member.utcOffset != nil ? "(UTC \(member.utcOffset ?? 0))" : ""
-        return "\(member.name ?? "") \(utcText)"
+        return "\(member.displayName()) \(utcText)"
```
rafaelks: 👍

Rocket.Chat/Views/ActionSheets/UserActionSheetPresenter.swift
```diff
@@ -0,0 +1,34 @@
...
+                controller.dismiss(animated: true, completion: nil)
+                self?.dismiss(animated: true, completion: nil)
+            }
+        }))
+
+        controller.addAction(UIAlertAction(title: localized("global.cancel"), style: .cancel))
+
+        present(controller, animated: true, completion: nil)
+    }
+}
```
rafaelks: 👍

rafaelks:
@cardoso I really liked the approach and I think this is going to be useful to our next release, where we're going to have more options on the user, like open the profile (in the future: mute, make admin, block, etc). 👍

cardoso:
Exactly @rafaelks 

I made this after I looked our issue with most the +1s : https://github.com/RocketChat/Rocket.Chat.iOS/issues/1319

That's probably coming next since we already have the API for that https://rocket.chat/docs/developer-guides/rest-api/channels/kick/#channel-kick

filipealva:
@cardoso This is a great approach! In the future, we can use a profile-like modally presented view controller instead of the action sheet, that way we will be able to provide more information about the user along with all the possible actions in a richer way.

PR #1610, opened 2018-04-27
[FIX] Custom emojis not updating when changing servers

cardoso:
@RocketChat/ios

I didn't notice any noticeable performance changes from this caching mechanism. If i'm wrong, please correct me. 

Otherwise LGTM.

### DIFF
--- a/Rocket.Chat/Managers/Model/CustomEmojiManager.swift
+++ b/Rocket.Chat/Managers/Model/CustomEmojiManager.swift
@@ -11,7 +11,10 @@ import RealmSwift
 
 struct CustomEmojiManager {
     static func sync() {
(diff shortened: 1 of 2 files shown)


rafaelks:
@cardoso Why don't we only clear the cache when changing server happens? This change will remove all of our caching on emojis 🤔

Rocket.Chat/Managers/Model/CustomEmojiManager.swift
```diff
@@ -11,7 +11,10 @@ import RealmSwift
 
 struct CustomEmojiManager {
     static func sync() {
+        CustomEmoji.cachedEmojis = nil
+
         let currentRealm = Realm.current
+
```
rafaelks: 👌

PR #1637, opened 2018-05-04
[NEW] Allow user with required permissions to remove member from room

### LINKED ISSUE #1319: [NEW] Allow user with permissions to remove users from a room

cardoso:
@RocketChat/ios

Also renames Subscription * Request -> Room * Request. It was starting to get confusing.

Adds test to room roles fetching

Closes #1319


rafaelks:
@cardoso I don't think I should be able to remove myself from the room 🤔:

simulator screen shot - iphone se - 2018-05-23 at 11 37 38
simulator screen shot - iphone se - 2018-05-23 at 11 37 31

cardoso:
@rafaelks do you think we should remove the option or rename it to "Exit Room" ?

rafaelks:
@cardoso Yes, let's not open anything to "myself" for now.

cardoso:
@rafaelks Made your changes and fixed conflicts!

Rocket.Chat/Controllers/Chat/MembersListViewController.swift
```diff
@@ -69,9 +69,21 @@ class MembersListViewController: BaseViewController {
     @IBOutlet weak var membersTableView: UITableView!
     var loaderCell: LoaderTableViewCell!
 
-    var data = MembersListViewData()
+    var data = MembersListViewData() {
+        didSet {
+            UIView.performWithoutAnimation {
+                membersTableView?.reloadData()
+            }
```
rafaelks: This is causing real weird behaviour on the list... we should be adding new rows instead of reloading the entire tableView. But not required right now. 👍

PR #1653, opened 2018-05-09
[IMPROVEMENT] Migrate subscriptions & rooms fetching to HTTP

### LINKED ISSUE #1517: [IMPROVEMENT] Move "rooms list" & "subscriptions list" call from WebSocket to REST
API to get rooms: https://rocket.chat/docs/developer-guides/rest-api/rooms/get/
API to get subscriptions: https://rocket.chat/docs/developer-guides/rest-api/subscriptions/get/

cardoso:
@RocketChat/ios

Closes #1517


cardoso:
Tests seem to be failing randomly. Trying to fix them.

cardoso:
Tests are ok now @rafaelks ✅

Rocket.Chat/Managers/Model/SubscriptionManager/SubscriptionManager.swift
```diff
@@ -22,121 +22,16 @@ struct SubscriptionManager {
...
-                        }
-                    }
-
-                    realm.add(subscriptions, update: true)
-                }, completion: {
-                    completion(response)
-                })
+        client?.fetchSubscriptions(updatedSince: lastUpdate) {
+            client?.fetchRooms(updatedSince: lastUpdate) {
+                completion?()
```
rafaelks: The only issue I see here is: if we've any error we're not continuing the app execution, right? Only if the error is the fallback version, of course. But what if by any reason one request fails with a 400 or 500? I think we can find a way to continue the execution here.

What do you think @cardoso?

rafaelks:
@cardoso This is looking great, I just made one comment. Let's discuss what's the best solution. 👍

Rocket.Chat/Managers/Model/SubscriptionManager/SubscriptionManager.swift
cardoso: I can think of a few options

1. Try again after some time (5s?)
2. Disconnect the user from the server
3. Continue execution despite outdated subscriptions/rooms

I'm torn between trying again and disconnecting the user from the server

rafaelks: @cardoso As we spoke:

1. Let's retry 3 times after 3 seconds on failure; After that, just ignore the error;
2. Always continue the flow, never logout the user;

Rocket.Chat/API/API.swift
```diff
@@ -41,6 +48,8 @@ final class API: APIFetcher {
     var authToken: String?
     var userId: String?
 
+    var retriesOnError: Int = 0
```
rafaelks: This is unused?

cardoso: Nice catch :+1:

PR #1657, opened 2018-05-10
[NEW] Add support to Broadcast rooms

### LINKED ISSUE #1654: [NEW] Support Broadcast
This feature allows the user to have channels/groups that are read-only with the possibility of other users being able to reply to the messages (in private).

Here’s how it works on the web:

*Left user is the admin and right user is the common user*

## What needs to be done

1. **Map the property broadcast in a channel/group**: now each channel has a property called *broadcast*. It is a boolean and the default value is **false**. The value can not be present and the app needs to handle as the default value;
2. **Members list must be hidden**: when the room have the **broadcast** property as **true**, the members list must be hidden. Only visible for admins, moderators and owners of the rooms;
3. **Normal users can’t send messages**: the room is read-only for normal users. If the client already supports read-only rooms, no changes are required.
4. **Reply button on each message**: each message in a room needs to have a **Reply** button and when the user taps on the button, it needs to open the DM with the user who sent the message and “quoting” the message link. Example: if the user taps on the reply button on message with identifier _abc123_ we need to open the DM with this text filled on the text view: `[ ](https://yourhost.com/direct/rafael.kellermann?msg=abc123)`.

## How’s the style of the Reply button

- Font: same on the messages list, with “semi-bold” or “medium”
- Font-size: 15pt
- Text color: #1D74F5
- Border color: #1D74F5
- Background color: #FFFFFF
- Border radius: 4px

rafaelks:
@RocketChat/ios

Closes #1654

## Progress

- [x] **Map the property broadcast in a channel/group:** now each channel has a property called broadcast. It is a boolean and the default value is false. The value can not be present and the app needs to handle as the default value;
- [x] **Members list must be hidden:** when the room have the broadcast property as true, the members list must be hidden. Only visible for admins, moderators and owners of the rooms; **(BLOCKED by https://github.com/RocketChat/Rocket.Chat/pull/10607)**
- [x] **Normal users can’t send messages:** the room is read-only for normal users. If the client already supports read-only rooms, no changes are required.
- [x] **Reply button on each message:** each message in a room needs to have a Reply button and when the user taps on the button, it needs to open the DM with the user who sent the message and “quoting” the message link. Example: if the user taps on the reply button on message with identifier abc123 we need to open the DM with this text filled on the text view: [ ](https://yourhost.com/direct/rafael.kellermann?msg=abc123).

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -10,8 +10,8 @@ import RealmSwift
 import SlackTextViewController
 import SimpleImageViewer
 
-private typealias NibCellIndentifier = (nibName: String, cellIdentifier: String)
-private let kEmptyCellIdentifier = "kEmptyCellIdentifier"
+typealias NibCellIndentifier = (nibName: String, cellIdentifier: String)
+let kEmptyCellIdentifier = "kEmptyCellIdentifier"
 
 // swiftlint:disable file_length type_body_length
 final class ChatViewController: SLKTextViewController {
--- a/Rocket.Chat/Controllers/Chat/ChatViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatViewController.swift
@@ -646,6 +646,7 @@ final class ChatViewController: SLKTextViewController {
             })
         }
 
+        updateSubscriptionRoles()
         updateMessageSendingPermission()
     }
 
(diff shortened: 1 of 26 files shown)


Rocket.Chat/Controllers/Chat/ChatViewController.swift
```diff
@@ -10,8 +10,8 @@ import RealmSwift
 import SlackTextViewController
 import SimpleImageViewer
 
-private typealias NibCellIndentifier = (nibName: String, cellIdentifier: String)
-private let kEmptyCellIdentifier = "kEmptyCellIdentifier"
```
rafaelks: Need to change this.

PR #1672, opened 2018-05-16
[FIX] Crashes on Authentication, subscription optional in Message & more

rafaelks:
@RocketChat/ios

# Description

- Changed the default thread of completion block on REST to the main thread;
- Subscription object is now optional on the Message, because sometimes we don't persist it (search, pinned, starred lists...);
- Fixed crashes on handling errors on authentication after the changes on REST;

### DIFF
--- a/Rocket.Chat/Controllers/Auth/AuthViewControllerAuthenticationHandler.swift
+++ b/Rocket.Chat/Controllers/Auth/AuthViewControllerAuthenticationHandler.swift
@@ -36,24 +36,26 @@ extension AuthViewController {
             }
         }
 
-        API.current()?.fetch(MeRequest()) { response in
+        API.current()?.fetch(MeRequest()) { [weak self] response in
             switch response {
             case .resource(let resource):
-                SocketManager.removeConnectionHandler(token: self.socketHandlerToken)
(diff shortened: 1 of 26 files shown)


Rocket.Chat/Controllers/Auth/AuthViewControllerAuthenticationHandler.swift
```diff
@@ -12,58 +12,48 @@ import RealmSwift
...
-                auth.serverVersion = version.description
+            if let realm = Realm.current, let auth = AuthManager.isAuthenticated(realm: realm), let version = serverVersion {
+                try? realm.write {
+                    auth.serverVersion = version.description
+                }
             }
         }
 
-        API.current()?.fetch(MeRequest()) { [weak self] response in
+        API.current()?.fetch(MeRequest()) { response in
```
cardoso: Why not weak self anymore? 🤔

rafaelks: Ups! Just fixed! 👍

rafaelks:
Gonna merge this one, the only test that's failing is on the timeout one...

PR #1687, opened 2018-05-21
[NEW] Add invite user to channel screen for users with permission to invite other users

### LINKED ISSUE #1686: [NEW] Allow user with permissions to invite other users to the room
- [ ] Button to add new members to the channel/group
  - [ ] List of all members [(API)](https://rocket.chat/docs/developer-guides/rest-api/miscellaneous/directory/)
  - [ ] Search on all members [(API)](https://rocket.chat/docs/developer-guides/rest-api/miscellaneous/directory/)
  - [ ] Add a new member to channel/group [(API)](https://rocket.chat/docs/developer-guides/rest-api/channels/invite/)

cardoso:
@RocketChat/ios

Closes #1686


Rocket.Chat/Models/Subscription/SubscriptionCanInviteUsers.swift
```diff
@@ -0,0 +1,31 @@
...
+
+        if type == .channel && user.hasPermission(.addUserToAnyChannelRoom, subscription: self) {
+            return true
+        }
+
+        if type == .group && user.hasPermission(.addUserToAnyPrivateRoom, subscription: self) {
+            return true
+        }
+
+        return false
```
rafaelks: As an admin, as the creator of the room, I was not able to invite users to the room... I think the logic here is not quite right. On the web, I can do it. Can you take a look please?

rafaelks (changes requested):

filipealva (commented):
Looking good! Just found an endless loading when asking for more pages with a searchText

filipealva:
After adding a user to a channel I'm getting stuck on a loading. The user only appears on the screen if I force a refresh using the pull to refresh.

Rocket.Chat/Models/Subscription/SubscriptionCanInviteUsers.swift
cardoso: Are you sure @rafaelks ? Can you provide steps to reproduce this? The permissions are working as expected for me.

Maybe there's some delay with fetching them. I'll do some tests.

rafaelks: @cardoso It worked for me now... I think was some kind of delay. 👍

filipealva (changes requested):
This is still happening:

filipealva:
@cardoso I found two more issues now:

1. If I open the members screen, then remove a member, then add it again, then pull to refresh so it can appear, then try to tap on him to show the action sheet with the options: nothing happens.

2. There's a crash, to reproduce:

- Open members screen
- Remove a member
- Add him again
- Click on him (nothing will happen as I described above, but it's required to reproduce the crash)
- Click on `back` 
- Close the `ChannelInfoViewController` and it crashes

cardoso:
@filipealva fixed that problem and made the screens work better with themes.

filipealva:
@cardoso great, the refreshing after adding issue is solved 💯  but the crash I've related on the last comment I did is still happening

rafaelks:
@cardoso This is looking super nice! I've couple of suggestions and two issues before we merge this:

### Issues

1. I should be able to add users to a public channel; Why is this not being allowed?
2. The offset of the members list is not right on opening it for a public channel (look at the GIF).



### Improvements

1. The list does not have a selected state; Let's add the default selection to the cell and only deselect on alert dismissal;
2. On the alert, you're hard coding the user's username; Let's use the `displayName()` method;

Rocket.Chat/Extensions/UITextField/UITextFieldActivityIndicator.swift
```diff
@@ -0,0 +1,44 @@
...
+        activityIndicator?.isHidden = false
+        clearButton?.isHidden = true
+    }
+
+    func stopIndicatingActivity() {
+        activityIndicator?.stopAnimating()
+        activityIndicator?.isHidden = true
+        clearButton?.isHidden = false
+    }
+}
```
rafaelks: 👍

rafaelks (approved):

rafaelks:
Very nice @cardoso, gonna merge it 👍

Rocket.Chat/Models/Subscription/SubscriptionCanInviteUsers.swift
saroar: > As an admin, as the creator of the room, I was not able to invite users to the room... I think the logic here is not quite right. On the web, I can do it. Can you take a look, please?

HI I have the same problem with who create a channel or group cant invite another user 
I think who create a group that person become the admin for that room so he should have access to add another person @rafaelks how its work my native IOS does not work

PR #1716, opened 2018-05-24
[CHORE] Replace SDWebImage with Nuke

### LINKED ISSUE #1578: [REFACTOR] Replace SDWebImage with Nuke
https://github.com/kean/Nuke

This could fix bugs like custom GIF emojis sometimes not animating. And it's one less Obj-C library.

filipealva:
@RocketChat/ios

Closes #1578


filipealva:
@cardoso Nuke relies on HTTP cache implementation of `Foundation.URLCache`, the issue here is that our server is returning `max-age=0` on cache control headers and `Foundation.URLCache` understands that it shouldn't cache it because of it. 

We have three options:

1 - Change it on the server
2 - Implement an aggressive on-disk caching layer ourselves (or use another lib just for caching) and easily integrate with Nuke
3 - Use Nuke's experimental aggressive on-disk caching feature

I would say that changing it on the server so we can rely on HTTP cache is a good option, but I don't know if it worth it to wait for it to merge this PR. 

Implementing a caching layer by ourselves or using another third-party library just for caching doesn't seem like a viable option, so I vote for using Nuke's experimental on-disk caching feature if we decide not to wait for the backend to update the cache control headers. 

What do you think about it?

cardoso:
@filipealva I think we should use Nuke's caching and ask for the back-end guys to implement the caching directives as soon as possible. I opened an issue here: https://github.com/RocketChat/Rocket.Chat/issues/10894

Let's test Nuke's aggressive caching. If it respects the Last-Modified header (if we're actually receiving it) for things like avatars and custom emojis then it should be ok.

rafaelks:
@filipealva Yes, unfortunately there's no way to cache the avatars today based on the HTTP cache layer, because of the point you mentioned. We can't afford waiting for a new release on the back-end to merge this one (only by 27th June) so let's go with the disk cache. Question: they only have an experimental caching layer, or they have something more robust and well-tested one?

filipealva:
@rafaelks The more robust well-tested approach for caching that they have is the HTTP Cache. This experimental aggressive caching layer is the only built-in alternative. I said "built-in" because it's easy to integrate a third-party caching library with Nuke, so this would be a way to go if we find out a well-tested caching library, but I think we shouldn't add one more dependency to do that.

PR #1729, opened 2018-05-29
[NEW] Autocomplete/Highlight should also happen for real names

### LINKED ISSUE #1595: [NEW] Autocomplete/Highlight should also happen for real names

filipealva:
@RocketChat/ios

Closes #1595


Rocket.Chat/Extensions/Models/UserExtensions.swift
```diff
@@ -21,17 +21,23 @@ extension User {
...
-            : realm.objects(User.self)).sorted(by: { user, _ in
-                guard let username = user.username else { return false }
-                return preference.contains(username)
-            })
+        let namePredicate = NSPredicate(format: "name CONTAINS[c] %@", word)
+        let usernamePredicate = NSPredicate(format: "username CONTAINS[c] %@", word)
+        let nameOrUsernamePredicate = NSCompoundPredicate(orPredicateWithSubpredicates: [namePredicate, usernamePredicate])
+        let roomMembersPredicate = NSPredicate(format: "username IN %@", preference)
+        let users = (word.count > 0 ? realm.objects(User.self).filter(nameOrUsernamePredicate)
+            : realm.objects(User.self).filter(roomMembersPredicate))
```
filipealva: @cardoso @rafaelks I thought it was a good idea to only show the room members when the user type just the `@`, this way the user can see only the most obvious choices for a mention. Then, when he starts typing we search for all the users (since it is possible to mention someone that is not in the room). 

What do you guys think about it? Should I keep showing users that aren't in the room when `word` is not `> 0` or my change make sense?

rafaelks:
Looking great!!! 👏

PR #1732, opened 2018-05-30
[FIX] Spotlight not working with new UI

cardoso:
@RocketChat/ios


cardoso:
Nice catch @rafaelks 😳 

Fixed the problem.

Rocket.Chat/Controllers/Subscriptions/SubscriptionsViewController.swift
```diff
@@ -68,6 +68,8 @@ final class SubscriptionsViewController: BaseViewController {
         updateData()
 
         SocketManager.addConnectionHandler(token: socketHandlerToken, handler: self)
+
+        tableView.reloadData()
```
rafaelks: Is this call really required? I think it's not.

cardoso: You're right 👍

PR #1787, opened 2018-06-18
[NEW] Translate ger 3.0

soundstorm:
Translation of german file; fixing some issues regarding different string lengths


kaiiiiiiiii:
Can you please use formal german instead of informal? The rest of the app is already in formal german to keep it in sync with the server core.

Thanks

PR #1796, opened 2018-06-20
[CHORE] Remove the resize of the image on upload

### LINKED ISSUE #1786: [IMPROVEMENT] Improve the logic on image compression
If the server maximum attachment size is very low, we need to improve the compression, and if the images do not have a limit on the size, we could send the original one.

rafaelks:
@RocketChat/ios

Closes #1786


cardoso:
Looks good @rafaelks . We also resize in the same way in the share extension. Do you want to change that too?

rafaelks:
Let's remove this one for now, we need to improve the logic here to support different styles of servers.

PR #1808, opened 2018-06-20
[I18N] Added Greek translations

kmav:
@RocketChat/ios

Added missing Greek translations. Also did a single ament in English translation of 
`"servers.action.disconnect.alert.message" = "Do you want to disconnect from \"%@\" server?";
`


kmav:
Thank you!

PR #1813, opened 2018-06-22
[IMPROVEMENT] Fix some memory related issues

### LINKED ISSUE #1801: [IMPROVEMENT] Resolve retain cycle on UIWindow's root view controller
After resolving the retain cycle on `SubscriptionsViewControllers` I found out that there's another happening on `BaseNavigationViewController` and `MainSplitViewController`

### LINKED ISSUE #1812: [BUG] Mistakenly switching selected server
1. Open the app with at least 2 servers authenticated
2. Expand servers list
3. Tap on `Add Server` 
4. Close it
5. Repeat the steps and you'll see that when you close each 2 times it changes the current selected server.

### LINKED ISSUE #1815: [IMPROVEMENT] Resolve API related memory leaks

filipealva:
@RocketChat/ios

### Highlight

We are getting rid of all these memory leaks with this PR:

screen shot 2018-06-22 at 12 09 03

Closes #1801
Closes #1815 
Closes #1812

### DIFF
--- a/Rocket.Chat/Extensions/API/APIExtensions.swift
+++ b/Rocket.Chat/Extensions/API/APIExtensions.swift
@@ -9,8 +9,6 @@
 import Foundation
 
 extension API {
-    static var current: API?
-
     static func current(auth: Auth? = AuthManager.isAuthenticated()) -> API? {
         guard
             let auth = auth,
@@ -19,21 +17,10 @@ extension API {
             return nil
         }
 
-        var api: API
-
-        if let current = current {
-            api = current
-        } else {
-            api = API(host: host, version: Version(auth.serverVersion) ?? .zero)
-        }
-
-        api.host = host
-        api.version = Version(auth.serverVersion) ?? .zero
+        let api = API(host: host, version: Version(auth.serverVersion) ?? .zero)
         api.userId = auth.userId
         api.authToken = auth.token
         api.language = AppManager.language
-
-        current = api
 
         return api
     }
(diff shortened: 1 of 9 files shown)


Rocket.Chat/Extensions/API/APIExtensions.swift
```diff
@@ -9,6 +9,8 @@
 import Foundation
 
 extension API {
+    static var current: API?
```
filipealva: @cardoso @rafaelks I added this property for us to reuse the current instance of `API`, it avoids having unnecessary instances in memory. Since we always call for `current` function below and it updates all the relevant information related to an `API` instance it seems that it will not cause any issue. 

Do you guys agree with this? Or am I missing some scenario where we can't share the same instance of an `API`? 


I'm asking it because I've found another way to solve the API related memory leaks, but when I remove this "single instance control" we wind up having always two different API instances in memory. They are not leaked because they're being referenced by `SocketManager.sharedInstace` which is always live throughout the application's lifecycle, but I think that we only need one instance.

rafaelks: @filipealva I believe it's fine for us to have a `current` property being shared all over the app, but I don't see any place you clearing this variable. Should we do that on changing servers at least?

filipealva: @rafaelks That's a good point. Actually, I think that this `current` can be a memory cache of our `API` instance throughout the application's lifecycle. It doesn't need to be cleaned because always when `current()` is called we refresh all of its properties, it should be a private (I would make it private on the next commit) property of `API` for memory caching.

I didn't think about it with the architecture point of view when I come up with this idea, I was just trying to isolate/solve the memory leaks. 

Maybe I just send the fix for the root memory leak and we take note of this multiple API instances and create a separate issue only for thinking about the best architectural approach, sounds good?

PR #1824, opened 2018-06-25
[CI] use correct gems versions

### LINKED ISSUE #1326: [CI] CI uses wrong version of CocoaPods
`[!] The version of CocoaPods used to generate the lockfile (1.4.0) is higher than the version of the current executable (1.3.1). Incompatibility issues may arise`

I think in `update_dependencies` script we should call `pod install` using bundler (`bundle exec pod install`) instead of calling it directly.

Side note: Do we have to call `bundle install` before `bundle exec`? What `before_install.sh` actually do? Why we do not keep `Gemfile.lock` file in the repository?

artrmz:
@RocketChat/ios

Closes #1326

Configured CI to install all gems using bundler. That will make sure we use proper gem's versions (based on Gemfile.lock). It's required to sync builds across all contributors and CI.
Before it used gems that were already installed on CircleCI and that could cause builds to fail if versions were not compatible.

In the future, if we want to upgrade any gem we should call `bundle update` and commit/push changes in `Gemfile.lock`. Also, it's preferable to use bundler locally while working on this project and use those commands:

To install gems:
`bundle install`

To use installed gems (ie. cocoapods):
`bundle exec pod install`


artrmz:
One test `SubscriptionsClientSpec.testFetchRoles()` seems to fail although there were no changes to actual code.

rafaelks:
@artrmz Looking nice! Thank you for that... just one question: do we need all this things installed? Why is the list so big, including things like `addressable`, `babosa`, etc?

rafaelks:
@artrmz I see! Makes sense, but we're not using Fastlane for now... can you remove it for now please?

artrmz:
@rafaelks As you wish, it's done 👍

PR #1825, opened 2018-06-25
[i18n] Add polish translation for new strings

artrmz:
@RocketChat/ios


rafaelks:
Great!!! Thank you @artrmz!

PR #1835, opened 2018-06-26
[IMPROVEMENT] Improve image compression

### LINKED ISSUE #1786: [IMPROVEMENT] Improve the logic on image compression

Sameesunkaria:
@RocketChat/ios

The image compression is not performed based on pre-calculated heuristics. An estimated size of the image after compression is determined based on the heuristics. This information is then used to calculate the new dimensions of the image so that it fits under the size constraints defined by `FileUpload_MaxFileSize`.

The heuristics were based on the stats below:

# Stats

## Percentage size vs compression quality
screen shot 2018-06-26 at 6 47 58 pm

## Percentage size vs percentage dimension
screen shot 2018-06-26 at 6 48 36 pm

Closes #1786


rafaelks:
@Sameesunkaria There's a test in `PublicSettingsRequestSpec` that needs to be updated.

rafaelks:
@Sameesunkaria Looking super good!!! :)

PR #1837, opened 2018-06-26
[NEW] User details screen

### LINKED ISSUE #1522: [NEW] Screen for "User" details

cardoso:
@RocketChat/ios 

Closes #1522


rafaelks:
@cardoso Looking really nice!

Two small things:

In landscape, the spacing between the final label and the screen it too short... let's add at least 16px (safe margin?)

simulator screen shot - iphone se - 2018-07-02 at 21 00 42

My timezone is UTC -3, not 09:01 `AM` but `PM` right now.

simulator screen shot - iphone se - 2018-07-02 at 21 01 14

cardoso:
Thanks for pointing these out @rafaelks . Fixed :+1:

Rocket.Chat/Controllers/Preferences/Profile/EditProfileViewModel.swift
```diff
@@ -23,15 +23,10 @@ final class EditProfileViewModel {
...
+            return localized("status.invisible")
         }
 
-        switch user.status {
-        case .online: return localized("status.online")
-        case .offline: return localized("status.invisible")
-        case .busy: return localized("status.busy")
-        case .away: return localized("status.away")
-        }
+        return user.status != .offline ? "\(user.status)" : localized("status.invisible")
```
rafaelks: Why this is not localized? 🤔

cardoso: @rafaelks it's localized in the description computed property :)

cardoso: But you reminded me that I forgot to localize some other stuff.

Sameesunkaria:
@cardoso Really liking how it turned out! 🎉

Just a few UI things, 

- The text items aren't positioned correctly when the view controller is coming on screen. (Might only be restricted to the iPhone X)

<p align=center>

</p>

- The nav bar remains transparent, when the message button is tapped and a new chat is opened.

<p align=center>

</p>

- The user details should be presented directly when in the members list view. And maybe an info ⓘ accessory be added to the cell.

<p align=center>

</p>

cardoso:
@Sameesunkaria I found that the first thing you mentioned only happens in the simulator. Can you confirm @rafaelks ?

rafaelks:
@cardoso @Sameesunkaria It happens on the device as well. 👍

filipealva:
@cardoso I had a similar issue with the navigation bar transition before and the following code solved it:

```
isNavigationBarHidden = true
isNavigationBarHidden = false
```

Adding it after you made the changes will force the navigation to redraw during the transition.

rafaelks:
Really nice @cardoso!

PR #1846, opened 2018-06-28
[FIX] Emoji avatars disappearing after a second

cardoso:
Not sure when this broke, but here's the fix.

@RocketChat/ios


artrmz:
Seems like a nasty thing that order of filling avatarView matters 🤔 Maybe actually should change the logic there? Feels like `updateAvatar()` might be called too many times, wouldn't it be better to hide those properties (username, emoji etc) and create a method for `AvatarView` filling that will contain proper logic and call `updateAvatar()` only once?

cardoso:
I agree @artrmz . AvatarView.swift deserves a refactor. I opened an issue for that and will take a look soon: https://github.com/RocketChat/Rocket.Chat.iOS/issues/1847

PR #1850, opened 2018-06-30
[IMPROVEMENT]  Allow setting theme properties through User Defined Runtime Attributes

### LINKED ISSUE #1845: [NEW] Allow user defined runtime attributes to define which color of the theme should be used
This is purely to facilitate development.

Add attributes that can be defined in interface builder:

(examples)
`themeColorType = 'auxiliaryText'`
`themeColorOverride = UIColor.red`

@Sameesunkaria

Sameesunkaria:
@RocketChat/ios

# Overview

This PR allows us to skip overriding the `applyTheme` method on a view, for theming trivial view attributes. Now basic theming can be done inside the Interface Builder, using User Defined Runtime Attributes.

These changes account for two different scenarios. 
- **CASE I**: If you want to set the color of an attribute to one of the colors available with the theme applied. (eg., `bodyText`)
- **CASE II**: If you want to set a color to an attribute, independent of what theme is currently applied. (eg., a label that stays red irrespective of the theme applied)

# Usage

On the interface builder, select the element you want to apply changes to. Under the `User Defined Runtime Attributes` section of the `Identity Inspector`, add the following key path with a value of type string

- **CASE I**: Key Path `themeColor`
- **CASE II**: Key Path `themeColorOverride`

### CASE I: Theme Color

The value for this key path is string containing the attribute of the view, which needs to be themed and the attribute of the theme which represents the desired color, separated with a colon.

```
textColor: auxiliaryText
```

This can be read as `view.textColor = theme.auxiliaryText`

<p align=center>
screen shot 2018-07-02 at 10 47 07 pm
</p>

### CASE II: Theme Color Override

The value for this key path is string containing the attribute of the view, which needs to be themed and the desired color represented in hex, separated with a colon.

```
backgroundColor: #ffffff
```

This can be read as `view.backgroundColor = UIColor(hex: "#ffffff")`

**TIP:** The color can be represented in various forms of hex, eg. `aaa`, `ababab`, `#ffffff`, `#AAAAAAFF`, etc. Some common color names are also supported, such as `red`, `green`, `blue`, etc. This behavior is adopted from the UIColor extension declared in the `UIColorCSSColorsExtension.swift` file.

<p align=center>
screen shot 2018-07-02 at 11 44 22 pm</p>

# Beyond the basics

`setThemeColor` and `setThemeColorOverride` are setters declared in `UIView`, which receive and store the values for the Runtime Attributes set in the Interface Builder. The method `applyThemeFromRuntimeAttributes` is responsible for applying these colors.

`applyThemeFromRuntimeAttributes` is already called in the base implementation of `applyTheme`, and is usually not required to be called again, if a view is being themed by overriding `applyTheme` so long as `super.applyTheme` is called.

But, in certain conditions it may be desirable to call `applyThemeFromRuntimeAttributes`, especially when overriding `applyTheme` for a generic view which may be subclassed in other places throughout the project. Here, it is recommended that `applyThemeFromRuntimeAttributes` be called at the end of the implementation of the overridden `applyTheme`. This allows for the attributes defined in the interface builder to take precedence over the code implemented in `applyTheme`.

## Type Safety

All the User Defined Runtime Attributes are stringly typed, so to make it easier to debug incorrectly entered attributes, assertions have been added so that the application will crash on debug builds (Release builds will not crash).

**Special thanks to @cardoso for his help on this PR. 🙌🏻**
Closes #1845


Rocket.Chat/Theme/ThemeableView.swift
```diff
@@ -0,0 +1,36 @@
+//
```
cardoso: This file can be removed, right?

Rocket.ChatTests/Managers/Launcher/AnalyticsCoordinatorSpec.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  BugTrackingCoordinatorSpec.swift
+//  AnalyticsCoordinatorSpec.swift
```
cardoso: Thanks :+1:

Sameesunkaria:
@cardoso That should be it.

PR #1879, opened 2018-07-04
[NEW] Support false "Store_Last_Message" setting & couple of bugs fixed

### LINKED ISSUE #1826: [IMPROVEMENT] Extended/Condensed channel view
With 3.0 we changed to the channel view mode called extended on web frontend. Before we had a condensed view with only the channel name and image. Some admins may have deactivated the option to store the last message, so users will see a list of
```
#channelname
No message

#channelname
No message

```
We may include a switch to collapse the view back to
```
#channelname
#channelname
```

rafaelks:
@RocketChat/ios

Closes #1826

This PR add support to the list of subscriptions when the setting *Store_Last_Message* is false (defaults to TRUE client-side). It also fixes the broken cell layout on loading new messages after entering the room.

In order to support this different layout, I created a brand new cell class to avoid constraints complexity and computer calculations that doesn't change on the same server. Let me know if you guys have a better idea please.

I also added another call to `rooms.get` and `subscriptions.get` because our API have issues with caching values and I think having a "give me all results" API may be safe to the client in order to have the whole data.


filipealva:
@rafaelks Is this PR supposed to fix this "cutting last message" bug? If so, it's still happenning

screen shot 2018-07-05 at 10 58 37

Rocket.Chat/Managers/Launcher/AnalyticsCoordinator.swift
```diff
@@ -44,9 +44,14 @@ struct AnalyticsCoordinator: LauncherProtocol {
     }
 
     private func launchFirebase() {
-        #if RELEASE || BETA
```
filipealva: @rafaelks wouldn't be better to leave it here to let it clear that we should only track events on these environments?

rafaelks:
@filipealva Changed! 👊

Rocket.Chat/Views/Cells/Subscription/SubscriptionCellCondensed.swift
```diff
@@ -0,0 +1,213 @@
+//
+//  SubscriptionCellCondensed.swift
+//  Rocket.Chat
+//
```
Sameesunkaria: @rafaelks This class shares a lot of code with the `SubscriptionsCell`, have you considered subclassing?

rafaelks: @Sameesunkaria Yes... we could have used subclass here maybe.

Sameesunkaria: @rafaelks How about we refactor it later?

rafaelks: @Sameesunkaria Yes, let's do it later in a refactor. 👍

PR #1887, opened 2018-07-05
[IMPROVEMENT] Reduce tappable area for sorting header view

Sameesunkaria:
@RocketChat/ios


filipealva:
@Sameesunkaria I've tested here. It works good, but I felt that we should put this non-tappable margin on the bottom of the tittle view instead of on the top of the sorting view.

filipealva:
@Sameesunkaria No, this kind of change can be done by editing the layout of the button on the xib file. 

But it's just a suggestion based on my feeling when interacting with this control. Let's see what @rafaelks and @cardoso think about this interaction before we decide to change anything again.

rafaelks:
@filipealva @Sameesunkaria What do you guys think about having 6px from the top for the sorting view and 6px from the bottom to the server button?

filipealva:
@rafaelks I think it's a good idea!

Rocket.Chat/Managers/AppManager.swift
```diff
@@ -155,6 +155,7 @@ extension AppManager {
         SocketManager.sharedInstance.connectionHandlers.removeAllObjects()
         SocketManager.disconnect { (_, _) in
             DispatchQueue.main.async {
+                UIApplication.shared.statusBarStyle = .default
```
filipealva: This line was removed recently because it does nothing since we manage the status bar style per view controller

Rocket.Chat/Managers/Socket/SocketManager.swift
```diff
@@ -164,36 +164,36 @@ extension SocketManager {
...
-                    PermissionManager.changes()
-                    infoClient.fetchPermissions()
-                    CustomEmojiManager.sync()
-
-                    commandsClient.fetchCommands()
-
-                    if let userIdentifier = auth.userId {
-                        PushManager.updateUser(userIdentifier)
-                    }
+            infoClient.fetchInfo()
```
filipealva: I'm not sure why these changes are being showed here 🤔 have you modified these files?

filipealva:
I've tested the interaction. It's looking good for me. Reduced the conflicts when interacting with the two controls.

Rocket.Chat/Managers/Socket/SocketManager.swift
Sameesunkaria: Oh, I merged the `bottom-spacing` branch and then reverted the changes. Maybe that is what's the issues.

PR #1935, opened 2018-07-10
[BUG] Attempt to fix a database inconsistency and some minor improvements on the server switching process

### LINKED ISSUE #1903: [BUG] Database inconsistency when signing up
I was testing the signup flow when I noticed that some rare times I was seeing subscriptions of the previous authenticated server on my new user subscriptions list. The references are invalid since the subscriptions don't actually exist for the server. 

I could record on the gif below one of the times it has happened. Notice the subscriptions `rafael.kellermann`, `filipe.brito` and `filipe.alvarenga` that keep on the screen when switching servers even though the account logged in `open.rocket.chat` should only have 2 subscriptions (the initial ones): general and sandbox. Another evidence is that we have `general` duplicated. 

We need to investigate this issue, it doens't happens all the time, but it's probaly a realm threading related issue.

filipealva:
@RocketChat/ios

This issue is very hard to reproduce. I'm not sure if this PR really fixes the issue or just make it harder to reproduce. We may need to rethink the entire way that we deal with the isolation of the databases. Just trusting in the "selected" doesn't work well when dealing with async calls (and we have A LOT of async calls that interact with the database on the callback). In the ideal world, we would have to think on how to make each call know which database/server it belongs to.

Closes #1903

### DIFF
--- a/Rocket.Chat/API/Clients/SubscriptionsClient.swift
+++ b/Rocket.Chat/API/Clients/SubscriptionsClient.swift
@@ -32,8 +32,6 @@ struct SubscriptionsClient: APIClient {
     func fetchSubscriptions(updatedSince: Date?, realm: Realm? = Realm.current, completion: (() -> Void)? = nil) {
         let req = SubscriptionsRequest(updatedSince: updatedSince)
 
-        let currentRealm = realm
-
         api.fetch(req, options: [.retryOnError(count: 3)]) { response in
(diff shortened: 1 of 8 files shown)


cardoso:
Sent a commit here to pass the parameter to the API.current call to avoid bugs in the future :top:

filipealva:
@cardoso good catch!

PR #1937, opened 2018-07-11
[CHORE] Many performance improvements & couple of improvements

rafaelks:
@RocketChat/ios

This PR improves the performance of the app by changing parts of the WebSocket message handling, simplifying the reutilisation of the SubscriptionCell and a couple of other small stuff. It also removes unused assets from the project and updates the Default icon asset.

Closes #1951 
Closes #1952

# Progress

- [x] Reduce 100% of the bottlenecks in the Subscriptions list;
- [x] Remove unused assets in the project (not related to performance);
- [x] Improve performance on avatarView reutilisation;
- [x] Start using the new button for "Scroll to Bottom" on the messages list, and improve it's performance;
- [x] Cache the cell height calculation for message cells;
- [x] Improve the performance of the messages list;

# To be done next, in a separated PR

- Reduce the amount of DB calls (Realm) on user status updates. There will be another way to fetch this data from the back-end in the future, but we've this problem right now and we need to deal with it.

PR #1943, opened 2018-07-11
[IMPROVEMENT] Theme scroll to bottom indicator

Sameesunkaria:
@RocketChat/ios

Depends on #1937


Sameesunkaria:
@filipealva an assets json file wasn't committed.

filipealva:
@Sameesunkaria just pulled your changes and it's still happening, only for the `dark` theme.

PR #1945, opened 2018-07-12
[FIX] Crash when trying to share a room on an iPad

### LINKED ISSUE #1925: [CRASH] Crash when trying to share a room on an iPad
```
Your application has presented a UIActivityViewController (<UIActivityViewController: 0x113295c00>). 
In its current trait environment, the modalPresentationStyle of a UIActivityViewController with this style 
is UIModalPresentationPopover. You must provide location information for this popover through the 
view controller's popoverPresentationController. You must provide either a sourceView and sourceRect 
or a barButtonItem. If this information is not known when you present the view controller, you may 
provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.
```

filipealva:
@RocketChat/ios

Closes #1925

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChannelActionsViewController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChannelActionsViewController.swift
@@ -225,9 +225,7 @@ extension ChannelActionsViewController {
             controller.popoverPresentationController?.sourceRect = shareRoomCell.bounds
         }
 
-        DispatchQueue.main.async {
-            self.present(controller, animated: true, completion: nil)
-        }
+        present(controller, animated: true, completion: nil)
     }
 
     override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Controllers/Chat/ChannelActionsViewController.swift
```diff
@@ -215,7 +218,16 @@ extension ChannelActionsViewController {
...
         let controller = UIActivityViewController(activityItems: [url], applicationActivities: nil)
-        present(controller, animated: true, completion: nil)
+
+        if shareRoomCell != nil && UIDevice.current.userInterfaceIdiom == .pad {
+            controller.modalPresentationStyle = .popover
+            controller.popoverPresentationController?.sourceView = shareRoomCell
+            controller.popoverPresentationController?.sourceRect = shareRoomCell.bounds
+        }
+
+        DispatchQueue.main.async {
```
rafaelks: Why are you opening another dispatch here?

filipealva: I noticed a delay between the tap on `share room` cell and the activity controller presentation. I thought it might be being called in background so I added this dispatch, and it has improved smoothness.

filipealva: I'm removing it since I double checked and it was already on main thread.

PR #1946, opened 2018-07-12
[FEATURE] Add Home Screen Quick Actions

### LINKED ISSUE #833: [NEW] Add Home Screen Quick Actions
Home screen quick actions. Customizable via settings if possible. Could add ie. most used chats (or favourites) as quick actions.

https://developer.apple.com/ios/human-interface-guidelines/extensions/home-screen-actions/

artrmz:
@RocketChat/ios

Closes #833

Added servers so far. Can add more types of shortcuts later on.

Do we want anything else except servers? Maybe channels and DMs? But since we are kinda limited with the number of shortcuts, we would need a good algorithm to determine which to shows - rather not the latest since it might be a little bit confusing (often) showing different shortcuts each time.

- [x] Dynamic items (servers)
- [x] Static Action - Add Server

### DIFF
--- a/Rocket.Chat/Managers/ShortcutsManager.swift
+++ b/Rocket.Chat/Managers/ShortcutsManager.swift
@@ -1,0 +1,49 @@
+//
+//  ShortcutsManager.swift
+//  Rocket.Chat
+//
+//  Created by Artur Rymarz on 12.07.2018.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+struct ShortcutServerItem {
+    let name: String
+    let url: String
+    let index: Int
+}
+
+struct ShortcutsManager {
+    static let serverIndex = "kServerIndex"
+
+    private static var shortcuts: [ShortcutServerItem]? {
+        let servers = DatabaseManager.servers?.enumerated().compactMap({ index, server in
+            ShortcutServerItem(name: server[ServerPersistKeys.serverName] ?? "", url: server[ServerPersistKeys.serverURL] ?? "", index: index)
+        })
+
+        return servers
+    }
+
+    static func sync() {
+        UIApplication.shared.shortcutItems = shortcuts?.compactMap({ shortcut in
+            let url = URL(string: shortcut.url)?.host ?? shortcut.url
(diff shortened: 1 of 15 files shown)


Sameesunkaria:
@artrmz Showing servers and DMs/channels together in one list can get confusing don't you think?

artrmz:
@Sameesunkaria Yea, that's why I went with just servers. Might also add one static action like "Connect to new server"? And if we want to directly open Channel/DM we could create a widget just like "Favourites" in Contacts?

Sameesunkaria:
@artrmz A widget for favorites is a really nice idea.

artrmz:
Added static action "Add Server".

Also noticed that `UsageDescription` keys are not localized. Will do that in another PR (already added `InfoPlist.strings` file in this PR)

Rocket.Chat/Managers/ShortcutsManager.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    static func sync() {
+        UIApplication.shared.shortcutItems = shortcuts?.compactMap({ shortcut in
+            let url = URL(string: shortcut.url)?.host ?? shortcut.url
+            return UIMutableApplicationShortcutItem(type: url, localizedTitle: shortcut.name, localizedSubtitle: url, icon: nil, userInfo: [serverIndex: shortcut.index])
+        })
+    }
+
+    static func selectServer(at index: Int) {
```
artrmz: `ServersListView` has pretty much the same method, should we extract it somewhere? Does it fit in `DatabaseManager` or `ServerManager`?

rafaelks:
Nice one @artrmz, thanks a lot! I think we could do like this: if user has more than 1 server, we show the servers, but if the user has only one server, we can show the most recent conversations?

artrmz:
@rafaelks Sure, will implement that tomorrow if I have time ;)

Rocket.Chat/Managers/ShortcutsManager.swift
rafaelks: Yeah... we could've use this one too maybe? `AppManager.changeSelectedServer()`. I'm going to merge it and then we can refactor in another PR, to avoid the conflicts for you.

PR #1968, opened 2018-07-13
[FEATURE] Add the latest rooms to quick actions when connected to the only server

artrmz:
@RocketChat/ios

Quick Actions on the home screen will now show servers to choose if we are connected to more than one server and if connected to only one server it will show latest rooms (currently showing 3 since there is one static action and system limits all actions to 4).


Rocket.Chat/Managers/AppManager.swift
```diff
@@ -110,6 +110,11 @@ struct AppManager {
 extension AppManager {
 
     static func changeSelectedServer(index: Int) {
+        guard index != DatabaseManager.selectedIndex else {
```
artrmz: Do we really have had all these things done while changing to the same server/index? Seems like `reloadApp` is just enough.

rafaelks:
Looking great, I'll just add a limit on the array processing to avoid processing data that won't be used! 👊

PR #1994, opened 2018-07-16
[FIX] Server isolation issues

### LINKED ISSUE #1991: [BUG] Presenting welcome view controller when canceling add server flow
We are facing a bug when canceling the action of adding a new server. To reproduce:

1. Have a server authenticated.
2. Tap on the server name to show the servers list
3. Tap on "Add new server"
4. Connect to the new server url
5. Tap on "back" to return to the server connection screen
6. Close the server connection screen
7. It will present the Welcome screen instead of getting back to the authenticated server

### LINKED ISSUE #1993: [BUG] Using another server cached settings
On the server switch flow we are using another server's cached settings until we fetch the current server settings again. It makes we setup a server based on the wrong settings then having to setup it again considering the correct settings

filipealva:
@RocketChat/ios

Closes #1991

Closes #1993

(Look at the last message, it shows when we use the wrong cached settings then it hides when we fetch the correct settings from the server)

### DIFF
--- a/Rocket.Chat/Managers/AppManager.swift
+++ b/Rocket.Chat/Managers/AppManager.swift
@@ -119,6 +119,7 @@ extension AppManager {
 
     static func changeSelectedServer(index: Int) {
         guard index != DatabaseManager.selectedIndex else {
+            DatabaseManager.changeDatabaseInstance(index: index)
             reloadApp()
             return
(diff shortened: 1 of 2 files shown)


rafaelks:
@filipealva Nice! 👍

PR #1988, opened 2018-07-16
[CRASH] Check indexPath of the cell before selecting it

rafaelks:
@RocketChat/ios

I know this crash is very weird, because the cell will be displayed, but here's the stack trace:

```
Fatal Exception: NSRangeException
*** -[__NSArrayM objectAtIndex:]: index 13 beyond bounds [0 .. 12]
```

```
Fatal Exception: NSRangeException
0  CoreFoundation                 0x18270ed8c __exceptionPreprocess
1  libobjc.A.dylib                0x1818c85ec objc_exception_throw
2  CoreFoundation                 0x1826a7750 _CFArgv
3  CoreFoundation                 0x1825d7810 -[__NSArrayM removeObjectAtIndex:]
4  UIKit                          0x18c592f18 -[UITableView _existingCellForRowAtIndexPath:]
5  UIKit                          0x18c5d1a7c -[UITableView _deselectRowAtIndexPath:animated:notifyDelegate:]
6  UIKit                          0x18c36e910 -[UITableView _deselectAllNonMultiSelectRowsAnimated:notifyDelegate:]
7  UIKit                          0x18c5e7d2c -[UITableView _selectRowAtIndexPath:animated:scrollPosition:notifyDelegate:]
8  Rocket.Chat                    0x10063af78 specialized SubscriptionsViewController.tableView(UITableView, willDisplay : UITableViewCell, forRowAt : IndexPath) -> () (SubscriptionsViewController.swift:418)
9  Rocket.Chat                    0x100638eb8 @objc SubscriptionsViewController.tableView(UITableView, willDisplay : UITableViewCell, forRowAt : IndexPath) -> () (SubscriptionsViewController.swift)
10 UIKit                          0x18c3ac2a0 -[UITableView _createPreparedCellForGlobalRow:withIndexPath:willDisplay:]
11 UIKit                          0x18c3abe00 -[UITableView _createPreparedCellForGlobalRow:willDisplay:]
12 UIKit                          0x18c3aaa68 -[UITableView _updateVisibleCellsNow:isRecursive:]
13 UIKit                          0x18c3a6668 -[UITableView layoutSubviews]
14 UIKit                          0x18c2e3770 -[UIView(CALayerDelegate) layoutSublayersOfLayer:]
15 QuartzCore                     0x18688525c -[CALayer layoutSublayers]
16 QuartzCore                     0x1868893ec CA::Layer::layout_if_needed(CA::Transaction*)
[...]

PR #2034, opened 2018-07-19
[FIX] Subscriptions should not repeat when they are grouped by type

### LINKED ISSUE #1982: [BUG] Only a single subscription cell is selected even if multiple for the same room appear on screen
Here, I tapped the blue cell but the red cell is the only one which is shown as selected.

This only applies to the iPad and iPhone Plus

img_0484

Sameesunkaria:
@RocketChat/ios

Closes #1982 
Closes #1988


rafaelks:
@Sameesunkaria We actually need to keep duplicated the ones that are unread. If you have unread in the top, they also need to be on the other groups as well. Can we do that?

Sameesunkaria:
@rafaelks The web client does not repeat any subscriptions, even if they are unread.

rafaelks:
@Sameesunkaria At some point they did, but that's true! So looks good! :-)

PR #2049, opened 2018-07-20
[BUG] Hide login button when setting is disabled

rafaelks:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
@@ -129,10 +129,9 @@ class AuthTableViewController: BaseTableViewController {
 
         guard let settings = serverPublicSettings else { return }
 
-        emailAuthRow.loginButton.isHidden = !settings.isUsernameEmailAuthenticationEnabled
-
         if !settings.isUsernameEmailAuthenticationEnabled {
             emailAuthRow.registerButton.isHidden = true
+            emailAuthRow.loginButton.isHidden = true
         } else {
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
```diff
@@ -124,10 +124,13 @@ class AuthTableViewController: BaseTableViewController {
...
         super.viewDidLoad()
+
         title = serverURL.host
 
         guard let settings = serverPublicSettings else { return }
 
+        emailAuthRow.loginButton.isHidden = !settings.isUsernameEmailAuthenticationEnabled
+
         if !settings.isUsernameEmailAuthenticationEnabled {
             emailAuthRow.registerButton.isHidden = true
```
filipealva: @rafaelks we could put the new line inside the `if` block. That way we would evaluate `settings.isUsernameAuthenticationEnabled` just once.

filipealva:
@rafaelks actually if `isUsernameAuthenticationEnabled` we shouldn't even show the `emailAuthRow`  and neither the separator of the login services section

rafaelks:
@filipealva Done!

filipealva:
@rafaelks Nice! but just hiding the cell still let us considering its height on uitableview's content size, which means that if you have a server with a certain number of login services you'll have an issue with an extra offset

rafaelks:
@filipealva I see... this is not using AutoLayout, so when the registerButton was hidden it was already having a white space. Will make some adjustments here.

PR #2051, opened 2018-07-20
[CHORE] Bouncing reactors & badge all the time for unreads

rafaelks:
@RocketChat/ios


filipealva:
A little description on the PR is good when there's no issue associated 🤓 

About the bouncing: awesome! 

About the unread badge for all unread messages: I like the idea but the `!` is so thin that it's annoying to look at, and I would easily confuse it with the number `1`. 

I tried letting the font `bold` when showing the `!` it looked a little bit better for me. Maybe we should proceed with the `!` bold for now and then work with Thiago to improve it.  What do you think about?

filipealva:
screen shot 2018-07-20 at 15 46 21


Just one more thing 🤓 we could increase the spacing between the label and the badge just a little bit.

PR #2056, opened 2018-07-20
[IMPROVEMENT] Improve Share Extension experience 

### LINKED ISSUE #1874: [BUGS] Share extension works weird
Any direction:
- [x] Clicking "Search" on keyboard clears searchbar and therefor resets the list
- [x] Cancel button is shown if tapped on searchbar but leaving it empty, then selecting channel and going back to overview (or simply by hitting "Search" like above)
- [x] At least with 3.0.0 I see duplicate channel names

Landscape (opening share menu in landscape, not switching from portrait to landscape while menu is opened):
- [x] Tapping searchbar doesn't open keyboard
- [x] When selecting channel/group/user the recipient hides under top bar when using the iOS screenshot instant share function (not if sharing from gallery) due to a not hidden statusbar

cardoso:
@RocketChat/ios 

Lots of things to improve here, code as well, I'll add them below as I finish them.

- [x] Don't clear search when search is pressed on keyboard
- [x] Don't show cancel button when search is empty after coming back from another screen
- [x] Fetch subscriptions remotely (no more empty list if local DB is empty)
- [x] Animate present & dismiss of Share Extension
- [x] Fix keyboard not showing if starting in landscape mode
- [x] Do not show favorite rooms in other sections too

Closes #1874

### DIFF
--- a/Rocket.Chat.ShareExtension/Base/SENavigationController.swift
+++ b/Rocket.Chat.ShareExtension/Base/SENavigationController.swift
@@ -37,7 +37,22 @@ final class SENavigationController: UINavigationController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+
+        view.transform = CGAffineTransform(translationX: 0, y: view.frame.size.height)
+
+        UIView.animate(withDuration: 0.25) { [weak self] in
+            self?.view.transform = CGAffineTransform.identity
+        }
+
         store.subscribe(self)
+    }
+
+    override func dismiss(animated flag: Bool, completion: (() -> Void)? = nil) {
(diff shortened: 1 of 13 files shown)


cardoso:
@filipealva Thanks for the suggestions! I'll create another issue with the improvements you mentioned, and we can focus on these in another PR.

rafaelks:
@filipealva Can you send your code changes that animates in another PR please? I'm gonna merge this one for now.

Sameesunkaria:
@filipealva Are we talking about this `UIViewAnimationOptions(rawValue: 7 << 16)` animation option?

And, instead of tapping on the table view, we should use interactive keyboard dismissal.

PR #2057, opened 2018-07-21
[DOCS] Update CONTRIBUTING.md

filipedelimabrito:
From the official CocoaPods site we update CocoaPods by simply installing the gem again.

Additionally we need to run the `pod install` command from the project directory, otherwise it can be show a `[!] No 'Podfile' found in the project directory.` message.


artrmz:
Actually I forgot to update this file in PR about bundler. If we want to be sure contributors use correct version we should update guide to use bundler.

artrmz:
Created another PR about Bundler #2065

PR #2065, opened 2018-07-24
[DOCS] Update CONTRIBUTING.md

artrmz:
Should use bundler to install dependencies

@RocketChat/ios

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,13 +31,17 @@ git clone https://github.com/RocketChat/Rocket.Chat.iOS
 ```
 2. Install [Bundler](https://bundler.io)
 ```
-gem install bundler (sudo might be required in some cases)
+sudo gem install bundler
+```
+To update _Bundler_ (when there is new version available)
+```
+sudo gem update bundler
 ```
 3. Install all gem dependencies
 ```
 bundle install (sudo might be required in some cases)
 ```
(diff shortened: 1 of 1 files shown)


rafaelks:
Looks good, thank you @artrmz!

filipedelimabrito:
@artrmz and @rafaelks What's the big difference by using `bundler` instead `CocoaPods`?

Let me show an example: I pulled the latests changes in this repo and wasn't be able to build the app anymore. Then I thought it was happening because the dependencies version was outdated. So I went to the current [CONTRIBUTING file](https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/CONTRIBUTING.md) but I don't like the idea (because I don't know what's the main advantage of doing so) of installing [Bundler](https://bundler.io/), then I went to the [previous version](https://github.com/RocketChat/Rocket.Chat.iOS/blob/90fa5a60f5bd41799f0a547ccd76f93d9a7677db/CONTRIBUTING.md) of this file and easily I was able to update the dependencies (using `pod install`) without the need to download `Bundler` (that's great) and then build the project again.

artrmz:
@filipedelimabrito Bundler is gems manager, it makes sure you have correct version of those installed.

Maybe it is easier for some people because they have already cocoapods installed but without bundler there is no way to force people to use correct version of gems (ie. cocoapods)

PR #2068, opened 2018-07-24
[IMPROVEMENT] Prevent fullscreen white image from obscuring close button & fix iPhone X layout

### LINKED ISSUE #1810: [IMPROVEMENT] Prevent fullscreen white image from obscuring close button
If you open a white picture, it gets enlarged and covers the whole screen. From a user perspective, it is not obvious how to close the picture preview. I found out that an user has to slide sideway to close the preview. This is not intuitive.

Example:

cardoso:
@RocketChat/ios 

- Adds a semi-transparent square behind close button
- Single tap shows/hides close button
- Adds safe area for better layout in iPhone X

Preview: https://imgur.com/nu0y2Fr

Work was done in this commit: https://github.com/cardoso/SimpleImageViewer/commit/8222c338de0f285ca0c2d556c5d8dedd4a365b52

Closes #1810

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -42,7 +42,7 @@ PODS:
   - RealmSwift (3.7.4):
     - Realm (= 3.7.4)
   - semver (1.1.0)
-  - SimpleImageViewer (1.1.1):
+  - SimpleImageViewer (1.2.0):
     - FLAnimatedImage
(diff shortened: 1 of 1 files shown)


Sameesunkaria:
How about we use a smaller circle instead of a square, while retaining the same tappable area?

PR #2087, opened 2018-07-27
[FIX] iPhone plus navigation weird on landscape

cardoso:
@RocketChat/ios


Rocket.Chat/Controllers/Base/MainSplitViewController.swift
```diff
@@ -59,6 +59,22 @@ final class MainSplitViewController: UISplitViewController {
...
+                        if let chat = detailViewController as? ChatViewController {
+                            if let subscription = chat.subscription {
+                                subscriptions.openChat(for: subscription)
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
Sameesunkaria: Woah 😱. How about using `&&` ?

cardoso: Oops. Did that for debugging and forgot to optimize.

PR #2088, opened 2018-07-27
[IMPROVEMENT] Improve Share Extension usability

### LINKED ISSUE #2084: [IMPROVEMENT] Improve Share Extension usability
Addition to #2056

I found just three points where we could improve for now:

If the user starts searching for something then clears the search bar the list should show the unfiltered subscriptions, but today the list is empty.

It would be nice to cancel the searching if the user taps on the tableview. Today there's no way to cancel the search without closing the entire Search Extension.

It would be nice to have a curve on the presenting and dismissing animations. I added the same curve we use for the ServersListView on my local branch and it looks good. If you agree with this change I can push my commit

Sameesunkaria:
@RocketChat/ios

- Remove warnings
- Use natural animation (`UIViewAnimationOptions(rawValue: 7 << 16)`)
- Support interactive keyboard dismissal
- Empty search bar does not filter results
- Tapping cancel on the search bar no longer dismisses the share extension 

Closes #2084

### DIFF
--- a/Rocket.Chat.ShareExtension/Rooms/SERoomsViewController.swift
+++ b/Rocket.Chat.ShareExtension/Rooms/SERoomsViewController.swift
@@ -63,6 +63,11 @@ extension SERoomsViewController: UISearchBarDelegate {
     }
 
     func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        guard !searchText.isEmpty else {
+            store.dispatch(.setSearchRooms(.none))
+            return
+        }
+
         store.dispatch(.setSearchRooms(.searching(searchText)))
     }
 }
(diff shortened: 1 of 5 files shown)


Rocket.Chat.ShareExtension/Rooms/SERoomsViewController.swift
```diff
@@ -61,6 +63,11 @@ extension SERoomsViewController: UISearchBarDelegate {
     }
 
     func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        guard !searchText.isEmpty else {
+            store.dispatch(.setSearchRooms(.none))
+            return
+        }
```
Sameesunkaria: @cardoso Opted to add this here, instead of in `SEStore`; because it improves readability...? 🤷🏻‍♂️

cardoso: This is fine too. :+1:

cardoso: This is fine too :+1:

PR #2092, opened 2018-07-28
[FIX] Filter hidden rooms from share extension

### LINKED ISSUE #2091: [BUG] Share Extension is showing hidden channels, groups and DMs

Sameesunkaria:
@RocketChat/ios

Closes #2091

### DIFF
--- a/Rocket.Chat.ShareExtension/State/SEState.swift
+++ b/Rocket.Chat.ShareExtension/State/SEState.swift
@@ -42,7 +42,7 @@ struct SEState {
         let displayedRooms: [Subscription]
         switch searchRooms {
         case .none:
-            displayedRooms = rooms
+            displayedRooms = rooms.filter { $0.open }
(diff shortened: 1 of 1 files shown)


Rocket.Chat.ShareExtension/State/SEState.swift
```diff
@@ -42,7 +42,7 @@ struct SEState {
         let displayedRooms: [Subscription]
         switch searchRooms {
         case .none:
-            displayedRooms = rooms
+            displayedRooms = rooms.filter { $0.open }
         case .searching(let search):
             let search = search.lowercased()
             displayedRooms = rooms.filter {
```
rafaelks: What about the `case .started:`? Do we need to add it here too?

Sameesunkaria: Actually, yeah.

PR #2106, opened 2018-08-03
[CHORE] Add events for "New Workspace" & "Status Update"

rafaelks:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Controllers/Auth/WelcomeViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/WelcomeViewController.swift
@@ -122,6 +122,8 @@ class WelcomeViewController: BaseViewController {
             return
         }
 
+        AnalyticsManager.log(event: .newWorkspace)
+
         let controller = SFSafariViewController(url: url)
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Controllers/Auth/WelcomeViewController.swift
```diff
@@ -122,6 +122,8 @@ class WelcomeViewController: BaseViewController {
             return
         }
 
+        AnalyticsManager.log(event: .newWorkspace)
```
filipealva: I think it's good to rename this event to `showNewWorkspace` or something like that, to make it clear that the user just viewed the "new workspace" screen. In the future, we can add parameters to the `go.rocket.chat` deeplinks to identify when the user just created a workspace and is coming back to the app. That way we will be able to generate a pipeline where this `showNewWorkspace` is when the user enters and the event triggered by the deeplink is the conversion.

rafaelks: Makes sense, just updated it! 👍

PR #2143, opened 2018-08-27
[FIX] Messages getting cut and scrolling

### LINKED ISSUE #2123: [BUG] Messages are cut and not displayed fully
- Rocket.Chat app version: 3.0.4 (206)
- Rocket.Chat server version: 0.65.1
- Device: iPhone 6, iOS 10.3.3

Often there's a situation that multi-line message is not displayed fully and last line is cut.

However, it is possible to scroll the text down within message boundaries and it is then seen that last line of text is was missing, but that makes first line of the multi-line message to disappear. It seems the height of message box boundaries is not calculated correctly as it is sometimes one line less than it should be.

cardoso:
@RocketChat/ios

Closes #2123


soundstorm:
This was all needed to fix it? 🤭

Sameesunkaria:
Do we know where the number 62 comes from?

cardoso:
@Sameesunkaria someone changed the spacing in the xib and forgot to change it there.

PR #2155, opened 2018-08-31
[CHORE] Use latest swift for all pods

Sameesunkaria:
@RocketChat/ios

Use Swift 4.1 for all pods by default, except for those which require swift 3 support.


Podfile
```diff
@@ -77,18 +77,19 @@ target 'Rocket.ChatTests' do
 end
 
 post_install do |installer|
-  swift4Targets = ['OAuthSwift', 'TagListView', 'SearchTextField', 'Nuke', 'Nuke-FLAnimatedImage-Plugin', 'DifferenceKit', 'SwipeCellKit']
+  swift3Targets = ['semver', 'SlackTextViewController', 'MobilePlayer', 'SimpleImageViewer', 'RCMarkdownParser', 'RealmSwift']
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
-      config.build_settings['SWIFT_VERSION'] = '3.1'
+      config.build_settings['SWIFT_VERSION'] = '4.1'
```
rafaelks: 4.2 to support Xcode 10 already @Sameesunkaria? :-)

Sameesunkaria: Umm, it'll take some time for the dependencies to update to the latest version of swift. There are some changes in UIKit introduced with Swift 4.2, so it's probably better for us to keep it at 4.1. 

We can easily change it to 4.2 down the line.

rafaelks:
Looks good @Sameesunkaria, thank you! :-)

PR #2182, opened 2018-09-10
[IMPROVEMENT] Remove defaults to open.rocket.chat & add a new placeholder to the connection field

rafaelks:
@RocketChat/ios

This was a decision we made internally for both iOS and Android to avoid users getting automatically connected to our community server. Now the user will be forced to enter an URL in order to connect.

### DIFF
--- a/Rocket.Chat/Resources/cs.lproj/Localizable.strings
+++ b/Rocket.Chat/Resources/cs.lproj/Localizable.strings
@@ -97,6 +97,7 @@
 // Socket Connection
 "connection.title" = "Sign in to your server"; // TODO
 "connection.button_connect" = "Connect"; // TODO
+"connection.server_url.placeholder" = "your-company.rocket.chat"; // TODO
 "connection.offline.banner.message" = "Jste offline, zkuste se znovu připojit";
 "connection.connecting.banner.message" = "Connecting..."; // TODO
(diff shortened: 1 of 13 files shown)


cardoso:
Now that the placeholder is no longer a valid URL, I think the Connect button should be disabled until the user starts typing.

rafaelks:
@cardoso That's a great suggestion, will work on that! :-)

Rocket.Chat/Resources/cs.lproj/Localizable.strings
```diff
@@ -97,6 +97,7 @@
 // Socket Connection
 "connection.title" = "Sign in to your server"; // TODO
 "connection.button_connect" = "Connect"; // TODO
+"connection.server_url.placeholder" = "your-company.rocket.chat"; // TODO
```
Sameesunkaria: Should the placeholder look like a url at all? How about something like "Address for Rocket.Chat server" (or any other natural language string).

rafaelks: @Sameesunkaria This was suggestion from @thiagosanchz... I think makes sense because you're telling the user how the entry should look like. 🤔

rafaelks:
@cardoso Updated with your requested changes.

PR #2195, opened 2018-09-18
[WIP][NEW] Remove SlackTextViewController 

### LINKED ISSUE #760: [REFACTOR] Remove SlackTextViewController dependency

filipealva:
@RocketChat/ios

This PR removes SlackTextViewController and introduces RocketChatViewController. 

We'll open new PRs based on this branch to replace the implementations based on SLKTextViewController. 

Closes #760


Rocket.Chat/Models/Message/UnmanagedMessage.swift
```diff
@@ -0,0 +1,53 @@
...
+    var managedObject: Message
+    var text: String
+    var attachments: [Attachment]
+    var user: UnmanagedUser?
+}
+
+extension UnmanagedMessage {
+    init(_ message: Message) {
+        managedObject = message
+        identifier = message.identifier ?? "" // FA NOTE: We must check if we have a valid identifier before calling this init. If a message doesn't have an identifier we should consider it invalid.
```
Sameesunkaria: We can maybe change the protocol to use an optional init..?

filipealva: @Sameesunkaria that may be a good solution. We just have to know if we need the `identifier` to be optional in the first place. If we make `identifier` a required (non-optional) property we won't need to make this init optional. 

I mean, if we have a valid `managed` message we should be able to create an `unmanaged` from it.

Sameesunkaria: @filipealva There are other use cases for it as well. We wouldn't have to check for if an object is invalidated, if we can just return nil on the unmanaged object.

filipealva: @Sameesunkaria it makes sense 👍

@rafaelks can you handle this change when working on the data controller?

rafaelks: Yes, implemented in my PR (already merged) 👍

rafaelks:
THIS IS GETTING REALLY NICE GUYS!!!

PR #2212, opened 2018-09-26
[WIP][NEW] Rewrite data controller to messages view model of MessagesViewController

rafaelks:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatDataController.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatDataController.swift
@@ -8,132 +8,86 @@
 
 import Foundation
 
-enum ChatDataType {
-    case daySeparator
-    case unreadSeparator
-    case message
-    case loader
-    case header
-}
-
-struct ChatData {
-    var identifier = String.random(10)
-    var type: ChatDataType = .message
-    var timestamp: Date
-    var indexPath: IndexPath!
-    var height: CGFloat?
-
-    // This is only used for messages
-    var message: Message?
-
-    // Initializers
-    init(type: ChatDataType, timestamp: Date) {
-        self.type = type
-        self.timestamp = timestamp
-    }
-}
-
 final class ChatDataController {
-    var messagesUsernames: Set<String> = []
 
-    var dataCellHeight: [String: CGFloat] = [:]
-    var data: [ChatData] = [] {
+    var data: [MessageSection] = [] {
         didSet {
-            messagesUsernames.removeAll()
-            messagesUsernames.formUnion(data.compactMap { $0.message?.validated()?.user?.validated()?.username })
+
         }
     }
 
-    var loadedAllMessages = false
-    lazy var lastSeen: Date = {
-        return Date()
-    }()
-    var unreadSeparator = false
-    var dismissUnreadSeparator = false
+    func updateData() {
 
-    @discardableResult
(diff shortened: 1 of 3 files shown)


Rocket.Chat/Controllers/Chat/MessagesSizingViewModel.swift
```diff
@@ -0,0 +1,36 @@
...
+//  MessagesSizingViewModel.swift
+//  Rocket.Chat
+//
+//  Created by Rafael Streit on 25/09/18.
+//  Copyright © 2018 Rocket.Chat. All rights reserved.
+//
+
+import Foundation
+
+final class MessagesSizingViewModel {
```
filipealva: TBH, I don't know if it should be a model (Manager) used by the ViewModel to process along with the other models that are necessary to bind the MessagesView or if it should actually be inside another ViewModel. Or even be a separated ViewModel as you placed it. 🤔 💭

Rocket.Chat/Controllers/Chat/MessagesViewController.swift
```diff
@@ -52,27 +49,44 @@ final class MessagesViewController: RocketChatViewController {
...
             sizingCell.prepareForReuse()
-            sizingCell.viewModel = viewModel
+            sizingCell.viewModel = sectionViewModel
             sizingCell.configure()
             sizingCell.setNeedsLayout()
             sizingCell.layoutIfNeeded()
+
             let size = sizingCell.contentView.systemLayoutSizeFitting(UIView.layoutFittingCompressedSize)
-            heightCache[viewModel.differenceIdentifier] = size.height
+            viewSizingModel.set(height: size.height, for: sectionViewModel.differenceIdentifier)
```
filipealva: To avoid referencing `UIScreen.main` here: 

```
return CGSize(width: UIScreen.main.bounds.width, height: height)
``` 

We could cache the `CGSize` instead of the `height`. We just need to use `NSValue(cgSize:_)` and adjust the `set(height:_, for:_)` method to support a CGSize.

Then to retrieve from the cache we just call `nsValue.cgSizeValue`. This is something I was going to do but since my caching mechanism was temporary I wind up not doing.

Rocket.Chat/Views/Chat/New Chat/Cells/DateSeparatorCell.swift
```diff
@@ -0,0 +1,62 @@
...
+        contentViewWidthConstraint = contentView.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width)
+        contentViewWidthConstraint.isActive = true
+    }
+
+    func configure() {
+        guard let viewModel = viewModel?.base as? DateSeparatorChatItem else {
+            return
+        }
+
+        let calendar = Calendar.current
```
filipealva: We should move any code about processing the model and making it ready to present to the `ChatItem` which is serving as a `viewModel` for us.

rafaelks: ✅

Rocket.Chat/Controllers/Chat/MessagesSizingViewModel.swift
rafaelks: ✅

PR #2215, opened 2018-09-28
[IMPROVEMENT] Sequential message

filipealva:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Controllers/Chat/ChatSections/MessageSection.swift
+++ b/Rocket.Chat/Controllers/Chat/ChatSections/MessageSection.swift
@@ -41,10 +41,17 @@ final class MessageSection: ChatSection {
             ).wrapped)
         }
 
-        cells.append(BasicMessageChatItem(
-            user: user,
-            message: object.message
-        ).wrapped)
+        if !object.isSequential {
+            cells.append(BasicMessageChatItem(
+                user: user,
+                message: object.message
+            ).wrapped)
+        } else {
+            cells.append(SequentialMessageChatItem(
+                user: user,
+                message: object.message
+            ).wrapped)
+        }
 
         return cells
     }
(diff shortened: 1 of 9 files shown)


Rocket.Chat/Controllers/Chat/MessagesViewController.swift
```diff
@@ -68,14 +87,11 @@ extension MessagesViewController {
...
         } else {
-            let sizingCell = BasicMessageCell.sizingCell
-            sizingCell.prepareForReuse()
-            sizingCell.viewModel = sectionViewModel
-            sizingCell.configure()
-            sizingCell.setNeedsLayout()
-            sizingCell.layoutIfNeeded()
-
-            let size = sizingCell.contentView.systemLayoutSizeFitting(UIView.layoutFittingCompressedSize)
+            guard let sizingCell = UINib(nibName: sectionViewModel.relatedReuseIdentifier, bundle: nil).instantiate() as? SizingCell else {
```
filipealva: SizingCell protocol was created in order to avoid having to check the specific object of the section to manually reference the static sizing cell of each type. 

The only con is: for this to work we *must* assign the cell's nib name to its `reuseIdentifier`.  Which is quite easy since we can use `String(describing:_)` method on the cell class, and instead of returning `.zero` we can throw a fatal error, that way we will make sure this will only happen at development time and we can set a message enforcing that the `reuseIdentifier` must be the same as the nib name. 


What do you guys think about this solution?

rafaelks: I think that makes perfect sense for now, thanks for letting us know. 👍

Rocket.Chat/Controllers/Chat/ChatSections/MessageSection.swift
```diff
@@ -56,6 +63,10 @@ final class MessageSection: ChatSection {
             cell.delegate = self
         }
 
+        if let cell = cell as? SequentialMessageCell {
+            cell.delegate = self
+        }
+
```
rafaelks: What about always having a delegate in the cells, or perhaps checking the negative states, like date separator and loader instead?

filipealva: @rafaelks I also don't like the way it is today, but until we use `ChatMessageProtocol` I think that it's best to leave this way.

In the near future, after we get everything working, I'm planning to split `ChatMessageCellProtocol` and its functionality.

PR #2224, opened 2018-09-28
[NEW] Support autocompletion in new composer

cardoso:
@RocketChat/ios

I had to refactor AvatarView to be programmatic so it can work with UserHintCell<AvatarView>


Rocket.Chat/Controllers/Chat/MessagesViewControllerComposerDelegate.swift
```diff
@@ -24,27 +25,102 @@ extension MessagesViewController: ComposerViewExpandedDelegate {
...
+            commands.forEach {
+                viewModel.hints.append($0.command)
+            }
+        } else if prefix == ":" {
+            let emojis = EmojiSearcher.standard.search(shortname: word.lowercased(), custom: CustomEmoji.emojis())
+
+            emojis.forEach {
+                viewModel.hints.append($0.suggestion)
+            }
+        }
```
rafaelks: What do you think about moving this method to the view model as well? Maybe we can have one view model just for the hints? 🤔

Rocket.Chat/Views/Avatar/AvatarView.swift
```diff
@@ -179,13 +210,12 @@ final class AvatarView: UIView {
...
 }
 
 // MARK: Themeable
 
 extension AvatarView {
     override func applyTheme() {
-        labelInitials?.textColor = .white
+        labelInitials.textColor = .white
     }
 }
```
rafaelks: 👌

Rocket.Chat/Controllers/Chat/MessagesViewControllerComposerDelegate.swift
cardoso: I saw that you guys are doing this pattern, so I did the same. I created a MessageComposerViewModel and put this logic there :+1:

Rocket.Chat/Views/Avatar/AvatarView.swift
rafaelks: I believe you need to call super here... don't you?

Rocket.Chat/Controllers/Chat/MessagesViewControllerComposerDelegate.swift
rafaelks: 👍

Rocket.Chat/Views/Avatar/AvatarView.swift
cardoso: Not needed in this particular case I think, but I can do it for consistency.

PR #2228, opened 2018-10-01
[NEW] Image component

rafaelks:
@RocketChat/ios


Rocket.Chat/Views/Chat/New Chat/ChatItems/ImageMessageChatItem.swift
```diff
@@ -0,0 +1,34 @@
...
+    var descriptionText: String?
+    var imageURL: URL?
+
+    var differenceIdentifier: String {
+        return identifier
+    }
+
+    func isContentEqual(to source: ImageMessageChatItem) -> Bool {
+        return
+            identifier == source.identifier &&
```
filipealva: I confirmed and we shouldn't check for identity here, just for content equality.

PR #2236, opened 2018-10-04
[NEW] Quote/Reply attachment

filipealva:
@RocketChat/ios


filipealva:
@rafaelks this is ready for review. The bug that's happening is not related to these changes. For some reason we are losing the ability to reload the cells that are first displayed on the screen, but if we scroll up and get back it works. It may have something to do with our data controller or even with `RocketChatViewController` itself.

PR #2243, opened 2018-10-09
[TESTS][NEW] Remove User relationship with Message, fixed all unit tests and remove test realm instance

rafaelks:
@RocketChat/ios


filipealva:
@rafaelks these changes are looking good, but I'm not sure if the Realm calls you moved to main thread are going to block the main thread in some cases. And the reason I'm not sure is because it's hard to test within the current stage of the RocketChatViewController migration. 

What do you think about it?

rafaelks:
@filipealva The only places I changed to use the mainThread block is where they were already using the current thread to execute by calling directly the `write` method from Realm instance. In tests, any execute method is executed on the main thread.

PR #2253, opened 2018-10-15
[IMPROVEMENT] DataController fetching partial database messages and code improvements

rafaelks:
@RocketChat/ios

This pull-request resolves couple of issues:

1. Load initial messages partially from database instead of all messages. Right now it's loading 30 messages every page;
2. It loads paginated result from database as well instead;

### DIFF
--- a/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
+++ b/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
@@ -196,8 +196,9 @@ final class MessagesViewModel {
             var sections: [AnyChatSection] = []
 
             var previousSection: AnyChatSection?
-//            let messages = subscription?.fetchMessages(40, lastMessageDate: nil)
-            messagesQuery.forEach({ (object) in
+            let messages = subscription?.fetchMessages(40, lastMessageDate: nil) ?? []
+
+            messages.forEach({ (object) in
                 if let section = section(for: object, previous: previousSection) {
                     sections.append(section)
                     previousSection = section
--- a/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
+++ b/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
@@ -188,81 +188,90 @@ final class MessagesViewModel {
     /**
      This method is called on every update the messagesQuery get from Realm.
     */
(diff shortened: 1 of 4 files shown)


Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
```diff
@@ -291,7 +301,7 @@ final class MessagesViewModel {
         guard let subscription = subscription?.validated() else { return }
         guard let subscriptionDetached = subscription.unmanaged else { return }
 
-        let messages = subscription.fetchMessages(50, lastMessageDate: oldestMessage)
+        let messages = subscription.fetchMessages(30, lastMessageDate: oldestMessage)
```
filipealva: The PR is looking really good! I can feel the improvement when opening chats, but it seems to have a little inconsistency. 

On the description you said that it is fetching a page of `40` messages, and here on the code I see `30`, but the real issue is that when debbuging each time `fetchMessages` gets called I see 60 sections/messages being inserted on our list, so it seems to be doubling up this page size right here.

rafaelks: @filipealva Just fixed the description of the PR to this number... where do you see 60? Can you point to the breakpoint you're using?

filipealva: @rafaelks Initially I added the breakpoint inside `RocketChatViewController`'s UIKit extension, inside the batch updates on line 45, then I print the `changeset.sectionInserted` property and it gives me `60` items, i.e. after doing the diffing there are 60 insertions. 

Now I went back on it and tried to spot the same before the diffing, and I could manage to see the method `func handle(insertions: [Int], on messagesQuery: Results<Message>)` handling `60` items regularly. 

It doesn't happens always, sometimes there are 20 insertions and another aleatory numbers, but if you keep testing you'll see that the most common number of insertions we have is `60` which is the double of the size we should be handling.

See the screenshot below:

screen shot 2018-10-15 at 16 12 28

rafaelks: @filipealva I see... you're seeing the observer from Realm, so new results in database would result in all these new values, because the request we're doing it for 60 new messages. Good finding, I'll change the parameter to 30 for now. 👍

PR #2259, opened 2018-10-20
[NEW] Read receipt

filipealva:
@RocketChat/ios

Note: Initially I was adding read receipts for all message components, that way we would see the read receipt checkmark always on the bottommost content that belongs to the message being "viewed". After discussing it on our daily call we thought it would be better to leave it only on the bottom of the basic message with headers and the sequential ones. The result is the following:

screen shot 2018-10-20 at 04 27 45

You can see that instead of adding the checkmark on the right of the URL preview, aligned to its bottom we added it on the sequential message, even though the checkmark is related to the URL preview as well. 

I'm not sure which path we should follow, but I would say that adding it to the bottommost piece of content gives more context on what is being marked as "viewed".


filipealva:
The tests were not broken by this PR, I would suggest to merge it and fix the tests on another one

filipealva:
🤣 🤣 🤣

PR #2260, opened 2018-10-23
[NEW] Broadcast reply/Message actions

filipealva:
@RocketChat/ios


rafaelks:
@filipealva Feature looks good, but the button should look like this:

screen shot 2018-10-23 at 06 51 47

filipealva:
@rafaelks you're right. done 👍

PR #2268, opened 2018-10-29
[IMPROVEMENT] Minor improvements to theming on RCVC

Sameesunkaria:
@RocketChat/ios

- Add MessagesViewController as an observer for theming
- Use colors without transparency to improve performance
- Rename brightBlue to actionTintColor so that it matches the naming convention

### DIFF
--- a/Rocket.Chat/Theme/Theme.swift
+++ b/Rocket.Chat/Theme/Theme.swift
@@ -1,5 +1,5 @@
 //
-//  Themeable.swift
+//  Theme.swift
 //  Rocket.Chat
 //
 //  Created by Samar Sunkaria on 3/25/18.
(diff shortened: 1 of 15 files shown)


Rocket.Chat/Theme/Theme.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  Themeable.swift
+//  Theme.swift
```
rafaelks: Nice catch! :-)

Rocket.Chat/Theme/Theme.swift
```diff
@@ -28,7 +28,7 @@ class Theme: NSObject {
     @objc let auxiliaryText: UIColor
     @objc let tintColor: UIColor
     @objc let auxiliaryTintColor: UIColor
-    @objc let brightBlue: UIColor
+    @objc let actionTintColor: UIColor
```
rafaelks: Indeed a better name! 👍

PR #2272, opened 2018-11-06
[FIX] Crash when tapping on collapsible quote

filipealva:
@RocketChat/ios


rafaelks:
@filipealva This is looking nice, but the default value is now that the view is uncollapsed (expanded). Can we do the inverse and be always collapsed by default instead?

PR #2295, opened 2018-11-09
[BUG][RCVC] Left and right edges on iPad need to increase to match the other elements in the controller

### LINKED ISSUE #2276: [BUG][RCVC] Left and right edges on iPad need to increase to match the other elements in the controller

filipealva:
@RocketChat/ios

Closes #2276


rafaelks:
@filipealva Referenced PR is merged already! 👍

PR #2299, opened 2018-11-09
[CHORE] Use dataSorted to calculate cell's size

filipealva:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
+++ b/Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
@@ -153,14 +153,14 @@ final class MessagesViewModel {
      */
     func item(for indexPath: IndexPath) -> AnyChatItem? {
         guard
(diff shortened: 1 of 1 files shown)


Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
```diff
@@ -153,14 +153,14 @@ final class MessagesViewModel {
      */
     func item(for indexPath: IndexPath) -> AnyChatItem? {
         guard
-            indexPath.section < dataNormalized.count,
-            indexPath.row < dataNormalized[indexPath.section].elements.count
+            indexPath.section < dataSorted.count,
+            indexPath.row < dataSorted[indexPath.section].viewModels().count
```
rafaelks: Let's request `viewModels()` just once and save it to get the right reference later?

filipealva: 👍

PR #2316, opened 2018-11-14
[FIX][RCVC] Reintroduce room join feature

cardoso:
@RocketChat/ios


Rocket.Chat/Controllers/Chat/MessagesViewControllerEmptyState.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  MessageViewControllerEmptyState.swift
+//  MessagesViewControllerEmptyState.swift
```
rafaelks: 👌

Rocket.Chat/Controllers/Chat/MessagesViewControllerJoining.swift
```diff
@@ -0,0 +1,67 @@
...
+        guard let auth = AuthManager.isAuthenticated() else { return }
+        guard let subscription = self.subscription else { return }
+
+        Realm.executeOnMainThread({ realm in
+            subscription.auth = auth
+            realm.add(subscription, update: true)
+        })
+
+        self.subscription = subscription
+        updateJoinedView()
```
rafaelks: This doesn't need to be called, will be called on the setter

Rocket.Chat/Managers/Model/SubscriptionManager/SubscriptionManager.swift
```diff
@@ -70,11 +70,20 @@ struct SubscriptionManager {
...
             currentRealm?.execute({ (realm) in
                 guard let auth = AuthManager.isAuthenticated(realm: realm), auth.serverURL == serverURL else { return }
-                let subscription = Subscription.getOrCreate(realm: realm, values: object, updates: { (object) in
-                    object?.auth = msg == "removed" ? nil : auth
-                })
 
-                realm.add(subscription, update: true)
+                let subscription: Subscription
+
+                if let rid = object["rid"].string, rid != "", let sub = Subscription.find(rid: rid) {
```
rafaelks: `rid.isEmpty` instead of `!= ""`?

rafaelks: This code could be a `guard` as well, instead of an `if`. I think makes more sense... what do you think?

PR #2333, opened 2018-11-20
[FIX] Include path in OAuth callback url building

### LINKED ISSUE #2321: [BUG] OAuth callback URL does not use the 'path' of the server URL
- Your Rocket.Chat app version: 3.1.1
- Your Rocket.Chat server version: 0.70.4
- Device (or Simulator) you're running with: iPhone X / iPhone XR

Hello, first of all, thanks for you hard work, the app has improved significantly since I've been using it and it became a central part of my day :).

I have a small issue with the OAuth login workflow: my server is configured on a subpath of my domain name (ex: "https://example.com/rocketchat"), therefore the OAuth callback url will be something like "https://example.com/rocketchat/_oauth/..." but the app uses "https://example.com/_oauth", discarding the subpath from the server url.

Looking at the source code, the issue seems to be in OAuthManager.swift in the function callbackUrl(). A fix that seems to work is:

```
    static func callbackUrl(for loginService: LoginService, server: URL) -> URL? {
        guard
            let host = server.host,
            let callbackPath = loginService.callbackPath ?? loginService.service
        else {
            return nil
        }

        let path = server.path

        var portString = ""
        if let port = server.port {
            portString = ":\(port)"
        }

        return URL(string: "https://\(host)\(portString)\(path)/_oauth/\(callbackPath)")
    }
```

I am no swift developer and I'm pretty sure you would find a more elegant way to handle the case, but I can make a pull request with this code if need be.

Regards,

cardoso:
@RocketChat/ios

Closes #2321

### DIFF
--- a/Rocket.ChatTests/Managers/OAuthManagerSpec.swift
+++ b/Rocket.ChatTests/Managers/OAuthManagerSpec.swift
@@ -40,25 +40,29 @@ class OAuthManagerSpec: XCTestCase {
         let loginService = LoginService()
         loginService.service = "github"
 
-        // without path
-
         let serverURL: URL! = URL(string: "https://open.rocket.chat")
         let expectedURL: URL! = URL(string: "https://open.rocket.chat/_oauth/github")
 
         XCTAssertEqual(OAuthManager.callbackUrl(for: loginService, server: serverURL), expectedURL, "callbackURL returns expected url")
+    }
 
-        // with path
+    func testCallbackUrlWithPath() {
+        let loginService = LoginService()
+        loginService.service = "github"
 
-        let serverURLWithPath: URL! = URL(string: "https://open.rocket.chat/path")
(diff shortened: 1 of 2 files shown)


Rocket.ChatTests/Managers/OAuthManagerSpec.swift
```diff
@@ -43,11 +43,37 @@ class OAuthManagerSpec: XCTestCase {
...
 
-        XCTAssertEqual(OAuthManager.callbackUrl(for: loginService, server: serverURL), expectedURL, "callbackURL returns expected url")
+        XCTAssertEqual(
+            OAuthManager.callbackUrl(for: loginService, server: serverURL),
+            expectedURL,
+            "callbackURL returns expected url"
+        )
+    }
+
+    func testCallbackUrlWithPath() {
```
rafaelks: ❤️

PR #2335, opened 2018-11-21
[CHORE] Only use unmanaged objects on transforming markdown cache

rafaelks:
@RocketChat/ios

### DIFF
--- a/Rocket.Chat/Models/Message/UnmanagedMessage.swift
+++ b/Rocket.Chat/Models/Message/UnmanagedMessage.swift
@@ -31,6 +31,14 @@ struct UnmanagedMessageReaction: Equatable {
     var usernames: [String]
 }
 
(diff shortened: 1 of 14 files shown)


Rocket.Chat/Models/Message/UnmanagedMessage.swift
```diff
@@ -177,13 +190,141 @@ extension UnmanagedMessage {
...
-            type == .url
+                type == .audio ||
+                type == .image ||
+                type == .video ||
+                type == .textAttachment ||
+                type == .url
         )
     }
+
+    // swiftlint:disable function_body_length cyclomatic_complexity
```
rafaelks: This methods are being duplicated for unmanaged objects... I believe we can remove them from the managed objects in the future.

PR #2387, opened 2018-11-30
[NEW] Add border to all new components attachments/boxes

rafaelks:
@RocketChat/ios


rafaelks:
Tests are passing, just iOS 12 on iPad Pro is failing on timeout ✅

PR #2405, opened 2018-12-06
[FIX] Pushing multiple chat controllers on navigation stack when lagging

cardoso:
@RocketChat/ios

This happens on iPhone when there's a lag and the user presses a room multiple times.


rafaelks:
Gonna merge this, tests are passing, just failing on timeout.

PR #2414, opened 2018-12-10
[CRASH] Fixed a crash with invalid Realm thread

rafaelks:
@RocketChat/ios


Rocket.Chat/Controllers/Chat/MessagesViewModel.swift
```diff
@@ -184,8 +184,10 @@ final class MessagesViewModel {
      messages of the subscription attached to the view model.
      */
     internal func unsubscribe(for subscription: Subscription) {
```
cardoso: subscription parameter isn't being used anymore? Should we remove it and also in the subscribe method?

PR #2428, opened 2018-12-13
[BUG][RCVC] Deleted messages are not removed from the list

### LINKED ISSUE #2383: [BUG][RCVC] Deleted messages are not removed from the list
Deleted messages are only removed after closing and opening the chat again.

filipealva:
@RocketChat/ios

Closes #2383

### DIFF
--- a/Rocket.Chat/API/Clients/MessagesClient.swift
+++ b/Rocket.Chat/API/Clients/MessagesClient.swift
@@ -123,11 +123,18 @@ struct MessagesClient: APIClient {
         api.fetch(DeleteMessageRequest(roomId: message.rid, msgId: id, asUser: asUser)) { response in
             switch response {
             case .resource(let resource):
-                guard let message = resource.message else {
+                guard let resourceMessage = resource.message else {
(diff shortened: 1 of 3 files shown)


Rocket.Chat/API/Clients/MessagesClient.swift
```diff
@@ -122,7 +122,21 @@ struct MessagesClient: APIClient {
...
+                }
+
+                let shouldKeepMessageLocally = AuthManager.isAuthenticated()?.settings?.messageShowDeletedStatus ?? true
+                if !shouldKeepMessageLocally {
+                    Realm.executeOnMainThread(realm: realm) { realm in
+                        realm.delete(message)
+                    }
+                }
+
+                if let unmanagedMessage = resourceMessage.unmanaged {
```
cardoso: Since the API doesn't return the type of the message, what do you think we set it manually = 'd' in case success = true ?

filipealva: Good catch! Just pushed the changes

PR #2432, opened 2018-12-14
[BLOCKED][IMPROVEMENT] Remove markdown from push notifications

### LINKED ISSUE #1563: [IMPROVEMENT] Remove Markdown tags from in some parts of the app
- [x] In-app notifications;
- [x] Reply/Quote pop-up;
- [x] Push notifications

cardoso:
@RocketChat/ios

Partially closes #1563 
Blocked by https://github.com/RocketChat/Rocket.Chat/issues/12956


Rocket.Chat/Managers/MarkdownManager.swift
```diff
@@ -104,6 +104,7 @@ fileprivate extension RCMarkdownParser {
         parser.strikeAttributes = [NSAttributedString.Key.strikethroughStyle.rawValue: NSNumber(value: NSUnderlineStyle.single.rawValue)]
         parser.linkAttributes = [NSAttributedString.Key.foregroundColor.rawValue: UIColor.darkGray]
 
+        #if !EXCLUDE_MARKDOWN_DOWNLOAD
```
rafaelks: I love the idea of using this kind of configuration per scheme, but what about something "pattern", like this?

> ROCKETCHAT_MARKDOWN_DOWNLOAD_MEDIA

This way we:

1. Make sure the name is actually managed by us;
2. It's a positive checking, not negative, will be like this:

```swift
#if ROCKETCHAT_MARKDOWN_DOWNLOAD_MEDIA
...
#endif
```

I think it brings more clarity to the code... what do you think?

Rocket.Chat/Extensions/UIColor/UIColorCSSColorsExtension.swift
```diff
@@ -308,3 +309,5 @@ public extension UIColor {
         "YELLOWGREEN" : "9ACD32"
     ]
 }
+
+// swiftlint:enable all
```
rafaelks: 👀

Rocket.Chat/Managers/MarkdownManager.swift
cardoso: 1. Agreed with namespacing the flag
2. If positive then we have to **remember** to add it for every build configuration in every target :/ this is very specific for the Notification Service Extension to avoid including all our download related classes.

PR #2442, opened 2018-12-21
[FIX][REGRESSION] Not sending typing event

### LINKED ISSUE #2440: [BUG] Not sending "typing" event

cardoso:
@RocketChat/ios

Closes #2440


cardoso:
@rafaelks I think this one is ready... But there's a delay in the server, we send the event and only a few seconds later the typing... appears.

rafaelks:
@cardoso From my tests, you're only calling the WebSocket call when user stops typing for a few seconds, can you check it again?

PR #2575, opened 2019-03-12
[FIX] Updated development environment setup instructions

### LINKED ISSUE #2553: Pod install make build not working
- My Rocket.Chat app version: 3.3.1
- Device (or Simulator) you're running with: Simulator

- I pull the newest version and build normal working.
- But i "pod install", this project has a problem.

rocketchatios issue

prateekguptaiiitk:
@RocketChat/ios

Closes #2553

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,16 +21,33 @@ In case you're interested in playing around with the code or giving something ba
 1. A macOS machine
 2. Xcode 9.3.x or higher (Swift 4.1)
 
-### Installing dependencies
+Before you begin, you should already have the Xcode downloaded and set up correctly. You can find a guide on how to do this here: [Setting up Xcode](https://developer.apple.com/xcode/)
 
-_(NOTE: Steps 3 and 4 must be executed from project's directory)_
+### Development Setup
 
-1. Clone this repo:
+##### &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; Steps to install Cocoapods (one time installation)
+
+- Run `sudo gem install cocoapods` to install the latest version of cocoapods. To install cocoapods from HomeBrew, `brew install cocoapods`.
+
+-  Next, run `pod setup` for setting up cocoapods master repo.
+
+### Setting up the iOS Project
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -21,16 +21,33 @@ In case you're interested in playing around with the code or giving something ba
...
+1. Download the _Rocket.Chat.iOS_ project source. You can do this either by forking and cloning the repository (recommended if you plan on pushing changes) or by downloading it as a ZIP file and extracting it. OR
 ```
-git clone https://github.com/RocketChat/Rocket.Chat.iOS
+$ git clone https://github.com/RocketChat/Rocket.Chat.iOS
 ```
-2. Use the Rocket.Chat.xcworkspace file to open the Xcode workspace.
-3. Build the project by ⌘ + R
+
+2. Navigate to the unzipped folder and run the following commands-</br>
+`pod deintegrate`</br>
```
rafaelks: @prateekguptaiiitk Why `deintegrate` is necessary here?

prateekguptaiiitk: `pod deintegrate` helps to remove all traces of Cocoapods from the project. So after that if we do `pod install` we can start to install them afresh. `pod update` will solve any version inconsistencies. Without `pod deintegrate` project was throwing random build errors.

rafaelks: @prateekguptaiiitk `pod update` will cause the CocoaPods to update libraries even if they're not in the same version as in `Podfile.lock`, right? It could cause some inconsistencies between projects in this case I think. Just `pod install` should be enough for anyone cloning the project, but most of the time that's not even necessary because the `Pods` folder is already being cloned.

Have you experienced any problems cloning the app for the first time and running it?

prateekguptaiiitk: Yes @rafaelks I along with some others experienced problems while setting up the project for the first time and running it. Even running `pod install` didn't help. When I ran `pod deintegrate` and then `pod install` my problem was resolved and project was built successfully.

snuff4: While I didn't experience any problem setting up the project for the first time, as the Pods were already there, I think the problem is usually with updating `swiftlint` and `fastlane`.

PR #2578, opened 2019-03-12
[FIX] Alert view for dark and black theme

### LINKED ISSUE #2577: [BUG] Alert not visible in Dark and Black Theme
- Your Rocket.Chat app version: 3.3.1
- Your Rocket.Chat server version: 1.0.0
- Device (or Simulator) you're running with: iPhone XR iOS 12.1.4 (Simulator)

The alertSuccess function creates a custom MBProgressHUD alert view, which is barely visible in the dark theme / not visible in black theme, in the following cases - 

- Save on notification preferences
- Updating profile
- Updating password
- Message being reported

Screenshots - 
Simulator Screen Shot - iPhone XR - 2019-03-12 at 23 33 08
Simulator Screen Shot - iPhone XR - 2019-03-12 at 23 22 58

snuff4:
@RocketChat/ios

Added an alert instead of MBProgressHUD custom alert view. It does not have the check image, but is consistent with the rest of the alert views in the application.

Closes #2577 

Screenshot - 

Simulator Screen Shot - iPhone XR - 2019-03-13 at 00 22 29

Simulator Screen Shot - iPhone XR - 2019-03-12 at 23 44 32

### DIFF
--- a/Rocket.Chat/Helpers/Alert.swift
+++ b/Rocket.Chat/Helpers/Alert.swift
@@ -42,11 +42,12 @@ extension UIViewController: Alerter {
     }
 
     func alertSuccess(title: String, completion: (() -> Void)? = nil) {
-        let alert = UIAlertController(title: title, message: nil, preferredStyle: .alert)
-        self.present(alert, animated: true, completion: nil)
-        let delay = DispatchTime.now() + 1.5
-        DispatchQueue.main.asyncAfter(deadline: delay) {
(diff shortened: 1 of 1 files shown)


rafaelks:
@snuff4 I'm afraid we don't want to use an alert on this screen... would you mind trying to change the theming for the `MBProgressHUD` instead? You can remove the theming from this view on the file Themeable.swift.

Thank you!

PR #2599, opened 2019-03-21
[IMPROVEMENT] Auto suggesting slash commands even when not first character

### LINKED ISSUE #2530: [IMPROVEMENT] Auto suggesting slash commands even when not first character
- Your Rocket.Chat app version: latest from test flight (234)
- Your Rocket.Chat server version: open
- Device (or Simulator) you're running with: xs

Seems to pop autocomplete even when typing / other places in the messages.  Kind of misleading since won’t work there

blewjy:
@RocketChat/ios

Closes #2530 

No hints will be generated if it is detected that the word containing the "/" character is not the first word.

Screenshots after fix:
not-first-word first-word


rafaelks:
Amazing, thank you!!! :)

PR #2602, opened 2019-03-22
[FIX] Clear listeners for dynamic value in notifications

artrmz:
Clear listeners for dynamic value in notifications before setting them again

Reloading cells caused multiple listeners to be registered which had a huge impact on performance and might cause other issues

@RocketChat/ios


rafaelks:
Great finding, thanks @artrmz!

PR #2608, opened 2019-03-25
[FIX] Refactor EmojiPickerViewController and fix memory leak on iPhone

### LINKED ISSUE #2423: [BUG] Retain cycle on EmojiPickerController
After picking an emoji `EmojiPickerController` is dismissed but we keep it holding the instance

artrmz:
@RocketChat/ios

Closes #2423 

I have refactored `EmojiPickerViewController` a little bit since there were a few issues with it (like calling a `super` of `loadView` - it should never be called while subclassing directly from `UIViewController`).

Moved presenter-related code to 'iPad only' since it was causing leaks on iPhones.

Please, double-check my changes since it was kinda tricky and not sure if it's the best solution.

NOTE: All tests succeed locally on my machine.

### DIFF
--- a/Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiPickerController.swift
+++ b/Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiPickerController.swift
@@ -13,52 +13,36 @@ final class EmojiPickerController: UIViewController, RCEmojiKitLocalizable {
     var emojiPicked: ((String) -> Void)?
     var customEmojis: [Emoji] = []
 
-    private var emojiPicker: EmojiPicker! {
-        didSet {
-            emojiPicker.emojiPicked = { emoji in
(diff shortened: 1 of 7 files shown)


Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiPickerController.swift
```diff
@@ -13,62 +13,50 @@ final class EmojiPickerController: UIViewController, RCEmojiKitLocalizable {
...
-        let center = NotificationCenter.default
-        center.removeObserver(self)
-        emojiPicker.endEditing(true)
+        super.viewWillDisappear(animated)
+
+        view.endEditing(true)
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
rafaelks: 💯

PR #2574, opened 2019-03-11
[CHORE] Update old libraries from Swift 3 to Swift 4.2

rafaelks:
@RocketChat/ios

PR #2616, opened 2019-03-29
[NEW] Authentication screen more accessible

snuff4:
@RocketChat/ios

TODO - 

- [x] Adding a comma between the accessibility label of "Join the community" and communityServerURL sounds a bit better by creating a pause in VoiceOver dictation.

- [x] The "More" right bar button on Login / Sign up to open Legal terms was not accessible, so I added localised text.

- [x] The cross button is described in VoiceOver as “Stop” button, which needs to be named as “Close” button.

- [x] Localised "Legal" Heading

- [x] "Show more options button" in Login should be accessible. (On the iPhone)

### DIFF
--- a/Rocket.Chat/Extensions/UIButton/UIButtonAccessibilityExtension.swift
+++ b/Rocket.Chat/Extensions/UIButton/UIButtonAccessibilityExtension.swift
@@ -1,0 +1,19 @@
+//
+//  UIButtonAccessibilityExtension.swift
+//  Rocket.Chat
+//
+//  Created by Rudrank Riyam on 11/04/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UIButton {
+    func applyShowMoreButtonAccessibility() {
+        accessibilityLabel = VOLocalizedString("auth.show_more_options.label")
+    }
+
+    func applyShowLessButtonAccessibility() {
+        accessibilityLabel = VOLocalizedString("auth.show_less_options.label")
+    }
+}
--- a/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
+++ b/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
@@ -9,7 +9,7 @@
 import UIKit
 
 extension UINavigationItem {
-    func moreButton() {
-        rightBarButtonItem?.accessibilityLabel = localized("auth.more")
+    func applyMoreButtonAccessibility() {
+        rightBarButtonItem?.accessibilityLabel = VOLocalizedString("auth.more.label")
     }
 }
--- a/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
+++ b/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
@@ -12,4 +12,8 @@ extension UINavigationItem {
     func applyMoreButtonAccessibility() {
         rightBarButtonItem?.accessibilityLabel = VOLocalizedString("auth.more.label")
     }
+
+    func applyCloseButtonAccessibility() {
+        leftBarButtonItem?.accessibilityLabel = VOLocalizedString("auth.close.label")
+    }
 }
--- a/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
+++ b/Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
@@ -1,0 +1,15 @@
+//
+//  UINavigationItem.swift
+//  Rocket.Chat
+//
+//  Created by Rudrank Riyam on 29/03/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UINavigationItem {
+    func moreButton() {
+        rightBarButtonItem?.accessibilityLabel = localized("auth.more")
+    }
+}
--- a/Rocket.Chat/Extensions/UINavigationBar/UINavigationItemAccessibilityExtension.swift
+++ b/Rocket.Chat/Extensions/UINavigationBar/UINavigationItemAccessibilityExtension.swift
@@ -1,5 +1,5 @@
 //
-//  UINavigationItem.swift
+//  UINavigationItemAccessibilityExtension.swift
 //  Rocket.Chat
 //
 //  Created by Rudrank Riyam on 29/03/19.
--- a/Rocket.Chat/Resources/cs.lproj/Localizable.strings
+++ b/Rocket.Chat/Resources/cs.lproj/Localizable.strings
@@ -114,8 +114,6 @@
 "auth.email_auth" = "e-mail"; // TODO
 "auth.login_service_prefix" = "Continue with "; // TODO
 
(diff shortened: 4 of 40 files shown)


Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Rudrank Riyam on 29/03/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UINavigationItem {
+    func moreButton() {
+        rightBarButtonItem?.accessibilityLabel = localized("auth.more")
```
cardoso: We have a separate localization method for accessibility:


https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.Chat/Helpers/VoiceOver/VOLocalized.swift

So this should be like 
```swift
VOLocalizedString("auth.more")
```

Rocket.Chat/Resources/cs.lproj/Localizable.strings
```diff
@@ -114,6 +114,8 @@
 "auth.email_auth" = "e-mail"; // TODO
 "auth.login_service_prefix" = "Continue with "; // TODO
 
+"auth.more" = "More"; // TODO
```
cardoso: Move these strings to VoiceOver.strings

https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.Chat/Resources/en.lproj/VoiceOver.strings

Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Rocket.Chat
+//
+//  Created by Rudrank Riyam on 29/03/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UINavigationItem {
+    func moreButton() {
```
cardoso: Can you think of a better descriptive name for this method?

Like `setupMoreButtonAccessibility()`

Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
snuff4: My apologies. I missed out on that. I have made the changes.

Rocket.Chat/Extensions/UINavigationBar/UINavigationItem.swift
snuff4: I changed it to `applyMoreButtonAccessibility()`

Rocket.Chat/Resources/cs.lproj/Localizable.strings
snuff4: Done. 👍

snuff4:
@rafaelks As I am using VoiceOver more, I am finding more accessibility labels that can be added. I will wrap up this PR by this week.

snuff4:
@rafaelks @cardoso The authentication screen is more friendly for VoiceOver users now . It would be awesome if you guys could try it out and give me some feedback on the implementation?

rafaelks:
@cardoso Most of the project is already updated, but you can contribute to my open PR if you want: https://github.com/RocketChat/Rocket.Chat.iOS/pull/2574. Thank you very much! :-)

snuff4:
@cardoso if the code is good, then it is good to merge. I have thoroughly tested in app, and have become a master of using VoiceOver in the past two months. 🥇

Rocket.Chat/Extensions/UIButton/UIButtonAccessibilityExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Rocket.Chat
+//
+//  Created by Rudrank Riyam on 11/04/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UIButton {
+    func applyShowMoreButtonAccessibility() {
```
Sameesunkaria: I believe this method does not belong here. We have to be careful while extending a class (especially the ones provided by apple). `applyShowMoreButtonAccessibility` is not something that we can apply to any generic `UIButton`, so adding this method as an extension to `UIButton` is not the best idea. 

You should maybe consider moving this method to the controller.

Rocket.Chat/Extensions/UIButton/UIButtonAccessibilityExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+        accessibilityLabel = VOLocalizedString("auth.show_more_options.label")
+    }
+
+    func applyShowLessButtonAccessibility() {
```
Sameesunkaria: same as above

Rocket.Chat/Extensions/UINavigationBar/UINavigationItemAccessibilityExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Rocket.Chat
+//
+//  Created by Rudrank Riyam on 29/03/19.
+//  Copyright © 2019 Rocket.Chat. All rights reserved.
+//
+
+import UIKit
+
+extension UINavigationItem {
+    func applyMoreButtonAccessibility() {
```
Sameesunkaria: Same concerns as for the `UIButton` thing, but since you are working with the nav bar item for the authentication controllers, I think you should work with the `AuthNavigationController`. That should allow you to do all your setup in one place.

Rocket.Chat/Extensions/UINavigationBar/UINavigationItemAccessibilityExtension.swift
```diff
@@ -0,0 +1,19 @@
...
+        rightBarButtonItem?.accessibilityLabel = VOLocalizedString("auth.more.label")
+    }
+
+    func applyCloseButtonAccessibility() {
```
Sameesunkaria: same as above

Rocket.Chat/Extensions/UIButton/UIButtonAccessibilityExtension.swift
snuff4: Thank you for your review. I understand your concern, and will work on it.
I have a solution in my mind right now, to replace the existing VOfunctions with a general method which takes the accessibilityLabel and returns the VOLocalizedString, and moving it to the controller

Rocket.Chat/Extensions/UINavigationBar/UINavigationItemAccessibilityExtension.swift
snuff4: Yes, creating an extension just for the accessibility is not a good idea. I will move the methods into the AuthNavigationController. 
Thank you!

rafaelks:
@snuff4 Are you working on the changes @Sameesunkaria requested you? Thanks!

snuff4:
@rafaelks yes yes, I am working on it. I found another bug where the VoiceOver is working outside the button also in the Login Screen, and working on a way to fix it.

PR #2642, opened 2019-04-04
[NEW] Crop image before uploading avatar

### LINKED ISSUE #2632: [NEW] Crop tool when uploading new avatar
When a user uploads a new avatar image right now, the app crops it into a maximum-sized square at the center of the image and sets it as the avatar automatically. It would be better if the user is presented with a crop tool that would allow them to select the portion of the image they would like to use as their avatar.

shloak17107:
@RocketChat/ios

Closes #2632 

Untitled 001

### DIFF
--- a/Rocket.Chat/Controllers/Preferences/Profile/EditProfileTableViewController.swift
+++ b/Rocket.Chat/Controllers/Preferences/Profile/EditProfileTableViewController.swift
@@ -525,7 +525,15 @@ extension EditProfileTableViewController: UIImagePickerControllerDelegate {
         let filename = String.random()
         var file: FileUpload?
 
-        if let image = info[.originalImage] as? UIImage {
+        if let image = info[.editedImage] as? UIImage {
+            file = UploadHelper.file(
+                for: image.compressedForUpload,
+                name: "\(filename.components(separatedBy: ".").first ?? "image").jpeg",
+                mimeType: "image/jpeg"
+            )
+
+            avatarView.imageView.image = image
+        } else if let image = info[.originalImage] as? UIImage {
             file = UploadHelper.file(
                 for: image.compressedForUpload,
                 name: "\(filename.components(separatedBy: ".").first ?? "image").jpeg",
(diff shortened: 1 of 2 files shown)


Rocket.Chat/Controllers/Preferences/Profile/EditProfileTableViewController.swift
```diff
@@ -525,7 +525,15 @@ extension EditProfileTableViewController: UIImagePickerControllerDelegate {
...
-        if let image = info[.originalImage] as? UIImage {
+        if let image = info[.editedImage] as? UIImage {
+            file = UploadHelper.file(
+                for: image.compressedForUpload,
+                name: "\(filename.components(separatedBy: ".").first ?? "image").jpeg",
+                mimeType: "image/jpeg"
+            )
+
+            avatarView.imageView.image = image
+        } else if let image = info[.originalImage] as? UIImage {
```
rafaelks: @shloak17107 It looks like the code here is the same as being used with the original image, do you think we can optimize it to only have it once?

Rocket.Chat/Extensions/UIViewController/UIViewControllerExtension.swift
```diff
@@ -34,7 +34,7 @@ extension MediaPicker where Self: UIViewController & UIImagePickerControllerDele
     func openPhotosLibrary() {
         let picker = UIImagePickerController()
         picker.delegate = self
-        picker.allowsEditing = false
+        picker.allowsEditing = true
```
rafaelks: 👍

Rocket.Chat/Controllers/Preferences/Profile/EditProfileTableViewController.swift
shloak17107: You are right. We can remove the else-if-condition where we are checking for the original image. This is because whenever the user selects an image for their avatar from the photo library, the cropping tool is always called. So, there will always be an .editedImage. 👍

PR #2679, opened 2019-04-23
[FIX] Prevent CustomEmoji image URL double url escaping

jungeonkim:
@RocketChat/ios


rafaelks:
@jungeonkim Are you seeing any problem that this change will fix it?

jungeonkim:
@rafaelks Yes. I tested it to work.
This problem is not a problem with custom emoji with names in common ASCII ranges, but custom emoji with URL escaped names like Korean(for example "웃음") does not currently shown properly.

rafaelks:
@jungeonkim Thank you! :-)

PR #2703, opened 2019-05-21
[NEW] Authentication screen VoiceOver accessible - Project 01

snuff4:
@RocketChat/ios

TODO - 

- [x] Adding a comma between the accessibility label of "Join the community" and communityServerURL sounds a bit better by creating a pause in VoiceOver dictation.

- [x] The "More" right bar button on Login / Sign up to open Legal terms was not accessible, so I added localised text.

- [x] The cross button is described in VoiceOver as “Stop” button, which needs to be named as “Close” button.

- [x] Localised "Legal" Heading

- [x] "Show more options button" in Login should be accessible. (On the iPhone)

### DIFF
--- a/Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
+++ b/Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
@@ -69,7 +69,7 @@ final class AuthTableViewController: BaseTableViewController {
         }
 
         collapsibleAuthSeparatorRow.showMoreButton.transform = CGAffineTransform(rotationAngle: CGFloat.pi)
-        collapsibleAuthSeparatorRow.showMoreButton.accessibilityLabel = VOLocalizedString("auth.show_more_options.label")
+        collapsibleAuthSeparatorRow.showMoreButton.accessibilityLabel = applyShowMoreButtonAccessibility()
         collapsibleAuthSeparatorRow.showOrHideLoginServices = { [weak self] in
             self?.showOrHideLoginServices()
         }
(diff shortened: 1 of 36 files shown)


Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
```diff
@@ -255,15 +266,18 @@ final class AuthTableViewController: BaseTableViewController {
 
         if isLoginServicesCollapsed {
             UIView.animate(withDuration: 0.5) {
-                self.collapsibleAuthSeparatorRow.showMoreButton.transform = CGAffineTransform(rotationAngle: CGFloat.pi)
+                self.collapsibleAuthSeparatorRow.showMoreButton.transform = CGAffineTransform(rotationAngle: .pi)
+                self.collapsibleAuthSeparatorRow.showMoreButton.accessibilityLabel =
```
cardoso: Why is this assignment inside animate?

Rocket.Chat/Controllers/Auth/AuthTableViewController.swift
```diff
@@ -255,15 +266,18 @@ final class AuthTableViewController: BaseTableViewController {
...
+                    self.showMoreButtonAccessibilityLabel
             }
 
             tableView.deleteRows(at: extraLoginServiceIndexPaths, with: .automatic)
         } else {
             UIView.animate(withDuration: 0.5) {
-                self.collapsibleAuthSeparatorRow.showMoreButton.transform = CGAffineTransform(rotationAngle: CGFloat.pi * 2)
+                self.collapsibleAuthSeparatorRow.showMoreButton.transform = CGAffineTransform(rotationAngle: .pi * 2)
+                self.collapsibleAuthSeparatorRow.showMoreButton.accessibilityLabel =
```
cardoso: Why is this assignment inside animate [2]? And since it's the same in `else` as in `if` you should probably put it outside and reduce it to only one assignment.

snuff4: I have updated the PR. Thank you for the review!

snuff4:
@Sameesunkaria why is the regex parsing test which is failing?

Sameesunkaria:
@snuff4 Its probably because of the compute power on CircleCI. All the tests pass on my Mac, so you can ignore that test for now.

snuff4:
All the tests pass on my Mac too. 
But I really don't like the way I am implementing the VOLocalizedStrings. 
I will add another PR with a better approach after I am done with the second project PR (which is very small)

cardoso:
@snuff4 if this is ready, can you remove the [WIP] flag and increase the failing test expectation from 5 seconds to 10?

snuff4:
@cardoso the test passed after increasing the timeout.

PR #2713, opened 2019-06-04
[I18N] German translation update 2019-05-06

AnBo83:
@RocketChat/ios


Rocket.Chat/Resources/de.lproj/Localizable.strings
```diff
@@ -4,7 +4,7 @@
 "global.no" = "Nein";
 "global.cancel" = "Abbrechen";
 "global.error" = "Fehler";
-
+chat.message.type.message_snippeted
```
cardoso: This doesn't look right... Can you remove this line and see if the tests pass?

AnBo83: Thank you for your reply, i've change the line.

PR #2740, opened 2019-07-18
[NEW] Channel icons VoiceOver accessible

rudrankriyam:
@RocketChat/ios

Added the VoiceOver accessibility for the channel icons.

### DIFF
--- a/Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconCell.swift
+++ b/Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconCell.swift
@@ -33,7 +33,7 @@ final class ChangeAppIconCell: UICollectionViewCell {
         } else {
             iconImageView.layer.borderColor = UIColor.RCLightGray().cgColor
             iconImageView.layer.borderWidth = 1
-            iconImageView.accessibilityTraits = .image
+            iconImageView.accessibilityTraits = .none
 
             checkImageView.isHidden = true
             checkImageViewBackground.isHidden = true
--- a/Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconCell.swift
+++ b/Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconCell.swift
@@ -10,16 +10,22 @@ import UIKit
 
 final class ChangeAppIconCell: UICollectionViewCell {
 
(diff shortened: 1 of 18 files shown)


Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconViewController.swift
```diff
@@ -61,7 +61,8 @@ extension ChangeAppIconViewController: UICollectionViewDataSource {
             fatalError("Could not dequeue reuable cell as ChangeAppIconCell")
         }
 
-        let iconName = viewModel.availableIcons[indexPath.row]
+        let iconName = Array(viewModel.availableIcons.keys)[indexPath.row]
+        let iconAccessibilityName = Array(viewModel.availableIcons.values)[indexPath.row]
```
Sameesunkaria: Maybe you could just get the value out by using the iconName as the index. Should be safer down the line. 

```swift
viewModel.availableIcons[iconName]
```

rudrankriyam:
I used a tuple now. Can you check? @Sameesunkaria

Rocket.Chat/Controllers/Preferences/ChangeAppIcon/ChangeAppIconCell.swift
```diff
@@ -10,23 +10,30 @@ import UIKit
...
             iconImageView.layer.borderWidth = 3
+            iconImageView.accessibilityTraits = .selected
 
             checkImageView.image = checkImageView.image?.imageWithTint(UIColor.RCBlue())
             checkImageView.isHidden = false
             checkImageViewBackground.isHidden = false
         } else {
             iconImageView.layer.borderColor = UIColor.RCLightGray().cgColor
             iconImageView.layer.borderWidth = 1
+            iconImageView.accessibilityTraits = .image
```
Sameesunkaria: @rudrankriyam I don't know if the icons should present themselves as an image. Maybe `.none` seems like a better fit. Lemme know how you feel about this.

rudrankriyam: .image was more precise, but as it is clear that it is an icon, .none fits better, especially with .selected. 
So, it is now -> selected white icon and white icon with .none instead of selected white icon and white icon image, with .image

Made the changes! @Sameesunkaria

PR #2742, opened 2019-07-22
[NEW] Reading / writing messages VoiceOver accessible - Project 06

rudrankriyam:
@RocketChat/ios

- [x] The text field in every message is spoken out as the particular line selected, ending with the first character of the next line. Each message as a whole should be accessible for dictation using VoiceOver, instead of a single line.
- [x] Clicking the Profile Picture to open the UserDetailView does not have any feedback.
- [x] “Upload media” button is not accessible. It is described in VoiceOver as “Button”, which should have an accessibilityLabel = “Share media”.
- [x] “Send” button is not accessible. It is described in VoiceOver as “Button”, which should have an accessibilityLabel = “Send”.
- [x] “Record” button is not accessible. It is described in VoiceOver as “Button”, which should have an accessibilityLabel = “Voice message”.
- [x] Localisation of Send and Record button accessibilityLabels.
- [x] AccessbilityLabel for composerView textfield and make the placeholderLabel inaccessible.

### DIFF
--- a/Pods/RocketChatViewController/Composer/Assets/cs.lproj/Localizable.strings
+++ b/Pods/RocketChatViewController/Composer/Assets/cs.lproj/Localizable.strings
@@ -1,3 +1,9 @@
 "composer.textview.placeholder" = "Type a message"; //TODO
 "composer.editingview.title" = "Editing Message"; //TODO
 "composer.recordaudioview.swipe" = "Swipe to cancel"; //TODO
+
+// Accessibility
+
+"composer.sendButton.label" = "Send"; // TODO
+"composer.micButton.label" = "Voice message"; // TODO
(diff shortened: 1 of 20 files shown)


Sameesunkaria:
The PR is looking good. Just a couple of things to that need to be fixed.

- The placeholder text is a label on top of the text field which is being read separately from the text field itself. Maybe consider hiding it, and add the placeholder text as a hint to the text field. (Take hints from the Messages app)
- The composer text view is just read out as "Text field", we should give it a proper label. 
-  When we are replying to a message, the accessibility of message preview (`ReplyView`) needs to be improved.
- The accessibility for the autocompletion view should be improved. (for both usernames (@) and emojis (:))

PR #2746, opened 2019-08-07
[NEW] Adding Reactions VoiceOver accessible - Project 08

rudrankriyam:
@RocketChat/ios

Closes #ISSUE_NUMBER

- [x] To open the Add Reactions View, double tap on the name and hold it for a second. (The Voice Over way of long press)
- [x] Categories.custom / categories.recent and all the other categories need to be assigned an accessibility label, which are defined in the RCEmojiKit.strings. Currently, they are described in VoiceOver as “Tab x of 10” button.
- [x] The skinTone button needs to be assigned accessibilityLabel = “Change Skin Tone” 
- [x] The custom emojis are not tapped as single elements, but as a whole view.
- [x] Add the labels for each custom emoji.

### DIFF
--- a/Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiCollectionViewCell.swift
+++ b/Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiCollectionViewCell.swift
@@ -22,6 +22,7 @@ final class EmojiCollectionViewCell: UICollectionViewCell {
         emojiLabel.baselineAdjustment = .alignCenters
         emojiLabel.font = UIFont.systemFont(ofSize: 32)
         emojiLabel.backgroundColor = UIColor.white
+        emojiLabel.isAccessibilityElement = true
         return emojiLabel
     }()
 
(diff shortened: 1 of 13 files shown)


Rocket.Chat/External/RCEmojiKit/Views/EmojiPicker/EmojiCollectionViewCell.swift
```diff
@@ -44,8 +44,11 @@ final class EmojiCollectionViewCell: UICollectionViewCell {
                 guard let url = url else { return }
                 ImageManager.loadImage(with: url, into: emojiImageView)
                 emojiImageView.isHidden = false
+                emojiImageView.isAccessibilityElement = true
+                emojiLabel.accessibilityElementsHidden = true
```
Sameesunkaria: You would wanna undo these changes in the .standard case. Because, the cells are reusable and later when the same cell is populated with a standard emoji, it'll be skipped by voiceover.

rudrankriyam: Ouch. I forgot this. I will fix this.

rudrankriyam:
Thank you for your review @Sameesunkaria !

And thanks for the skin tone names! I will edit the skinTone array from     
`let skinTones: [(name: String?, color: UIColor)]`
to 
`let skinTones: [(name: String?, color: UIColor, accessibilityName: String?)]`

For the default skinTone, I don't know why nil was used for the name. I will edit the accessibilityName to default.

PR #2748, opened 2019-08-13
[WIP] [NEW] Recording and sending audio VoiceOver accessible - Project 09

rudrankriyam:
@RocketChat/ios

- [x] The micButton is described in VoiceOver as “Mic Button” which needs to be assigned an accessibilityLabel = “Record Audio” 
- [x] The sendButton is described in VoiceOver as “Red Mic Button” which needs to be assigned an accessibilityLabel = “Stop recording” 
- [ ] The timeLabel is described in VoiceOver as “zero three” which needs to be accessible as “Duration three seconds.” This will be accomplished using DateComponentsFormatter.
- [x] “Swipe to cancel” button will be enabled for a left swipe using UIAccessibilityTraitAllowsDirectInteraction . 

The following buttons are described in VoiceOver with a redundant “Button” 
- [x] The playButton which needs to be assigned an accessibilityLabel = “Play” instead of “Play Button” ✅
- [x] The sendButton which needs to be assigned an accessibilityLabel = “Send” instead of “Send Button” ✅
- [x] The discardButton which needs to be assigned an accessibilityLabel = “Discard” instead of “Discard Button” ✅
- [x] The sendButton sometimes is described in VoiceOver as “Red Mic Button” ✅
- [ ] The timeLabel is described in VoiceOver as “zero two” which needs to be accessible as “Duration two seconds.” This will be accomplished using DateComponentsFormatter.
- [ ] Currently the Slider is described in VoiceOver as “Zero percent. Adjustable. Swipe up or down with one finger to adjust the value”.  It should be rather “Recording position zero seconds of two seconds. Adjustable.”
- [ ] Setting UIAccessibilityTraits.adjustable for the AudioView.
- [x] The playButton is described in VoiceOver as “Player Play” which needs to be assigned an accessibilityLabel = “Play Audio” ✅

I hid the accessibilityElements wherever applicable so it provides a smooth VoiceOver experience.

### DIFF
--- a/Pods/RocketChatViewController/Composer/Assets/cs.lproj/Localizable.strings
+++ b/Pods/RocketChatViewController/Composer/Assets/cs.lproj/Localizable.strings
@@ -11,3 +11,4 @@
 "composer.playButton.label" = "Play"; // TODO
 "composer.pauseButton.label" = "Pause"; // TODO
 "composer.discardButton.label" = "Discard"; // TODO
+"composer.duration.label" = "Duration,"; // TODO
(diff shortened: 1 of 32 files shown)


cardoso:
Awesome work. Thank you!

PR #2783, opened 2019-11-09
Upgrade to Xcode 11, update dependencies and fix a few resulting bugs

cardoso:
- Update project and dependencies to Xcode 11 and iOS 13 and fix various resulting bugs

Detailed list of changes:

- Update Realm
- Update JitsiMeet

- Fix black screen after login (iOS 13)
- Fix alerts disappearing right after presenting (iOS 13)
- Fix crash on In-App notifications (iOS 13)

- Fix Jitsi not working (white screen)
- Fix white screen on leaving Jitsi call
- Fix Jitsi screen on iPad being dismissable and call staying on


Rocket.Chat/Extensions/UIWindowExtensions.swift
```diff
@@ -41,9 +41,13 @@ extension UIWindow {
         }
 
         /// The presenting view controllers view doesn't get removed from the window as its currently transistioning and presenting a view controller
```
Sameesunkaria: How does this issue materialize in our app? Like what are the steps to reproduce it?

I was under the impression that this problem only exist if the dismiss method is not called on the parent vc.

Rocket.Chat/Extensions/UISearchBar/UISearchBarTextField.swift
```diff
@@ -10,6 +10,10 @@ import UIKit
 
 extension UISearchBar {
     var textField: UITextField? {
+        if #available(iOS 13.0, *) {
+            return searchTextField
```
Sameesunkaria: Finally! 🙌🏻

